PR #3, opened 2015-04-12
Better coverage of mul high instructions

joerchan:
My riscv implementation passed both mulh and mulhu without me having implemented signed behaviour in my multiplication unit.


aswaterman:
This is a good thing to improve.  Alas, the new mulh and mulhsu tests don't seem to pass on spike.

joerchan:
Hi. Yes, you are right. Turns out I had a little too much faith in my own implementation of the multiplication unit. I had a bug which I fixed.
Now spike and my implementation agree on the result.

aswaterman:
Great, thanks for the contribution.

PR #6, opened 2016-01-28
Support make from this decade (newer than 3.81)

palmer-dabbelt:
This is a kind of roundabout fix: I can't run "make run-asm-tests" in
rocket-chip, and I believe the reason this is happening is because make-3.82
changed pattern rule matching to by "sort by shortest" as opposed to "sort by
first definition".  I've been unable to verify this, but make-3.82 is trying to
match "mm" when building "mm.riscv.hex" while make-3.81 is trying to match
"mm.riscv" when trying to build "mm.riscv.hex".

Since I can't find a real fix, I've just put every benchmark in a ".d"
directory, which is the standard fix when you can't figure out why there's a
file/directory namespace collision.

I've tried this for me on make-4.1 and make-3.81, but it's so fragile it
deserves further testing.


palmer-dabbelt:
@aswaterman @colinschmidt This works for me, but can you guys test it?  make-3.81 is 10 years old.

aswaterman:
Why not just get rid of the namespace collision?

palmer-dabbelt:
That's probably the right fix, but I tried for an hour or so and couldn't figure it out.  3.82 is in ~palmer.dabbelt/.local/src/make on most machines, if you want to try.

IIRC you said 3.82 didn't work for you for some reason?  Do you remember why?  That might help me figure out what's going on.

aswaterman:
It failed to build something important. I think it was the gcc toolchain,
not the tests. There is a reason that Ubuntu is on a 10 year old version of
make.

PR #7, opened 2016-01-30
Two extra NaN test cases

palmer-dabbelt:
I've been experiencing some differences between Spike and Rocket when trying to run riscv-torture that look like they're related to NaNs.  These test cases add a few arbitrary NaNs to the ISA tests that I happened to write when trying to figure out what was going on.


palmer-dabbelt:
@aswaterman There's a second test, but that one is a lot simpler.

PR #8, opened 2016-02-18
What is the square root of 171?

palmer-dabbelt:
I get differnt answers on Rocket and Spike.  The Spike answer matches my Intel
machine.


palmer-dabbelt:
@aswaterman I don't know anything about hardfloat or if Rocket is expected to match Spike for this.  I managed to get the hardfloat repo to fail a test case with the following input
````
palmer.dabbelt a5 berkeley-hardfloat $ head -n1 log
4065600000000000 4065600000000000 10
palmer.dabbelt a5 berkeley-hardfloat $ cat log | ./test-DivSqrtRecF64_sqrt/dut 0
[0000000] 4065600000000000
        => 0803a0d10482f43ae 01   expected: 08075600000000000 10
````
but I have no idea what the format should be so I'm not sure that's a valid complaint.

palmer-dabbelt:
I found a calculator and it agrees with Spike.

aswaterman:
Nice, you should bug JohnH about it.  Of course, it's possible that it's a Chisel bug, too.

palmer-dabbelt:
OK, as long as this is an actual bug (and not an expected failure) I'll take some time to investigate what's actually going on.

PR #10, opened 2016-02-27
Support multi configuration travis builds

colinschmidt:
This allows you to build the tests with

riscv{32,64}-unknown-{linux-gnu,elf}-gcc


palmer-dabbelt:
@colinschmidt This configure diff is unreadable, did you use "autoreconf -i" to produce it?  If so then it's probably just the version bump, but I'm not going to look at it (configure.ac looks fine).

colinschmidt:
Yes. I did

PR #11, opened 2016-02-29
Tests can now build with non RVG toolchains

colinschmidt:
This adds flags to disable M,A,F/D to accomplish this cleanly we separate the ui folders into um and ua.

This is needed for riscv/riscv-tools/pull/26. Assuming that PR passes [travis](https://travis-ci.org/riscv/riscv-tools/builds/112663064) I'm ready to merge this. I locally tested it with RV{32,64}{G,I} and the build succeeds and produces the expected tests.


aswaterman:
This looks good, but my preference would be to hold off on this until the priv-1.9 stuff is merged, as it will cause a significant merge conflict.  Is that okay, or will it cause you problems?

colinschmidt:
I think if I wait on this and disable the RVI test in travis it should be ok for the main pr in tools to go through. It would be good for me to get the PR off my mind, and I'm ok waiting on this one.

aswaterman:
OK.  In that case, let's revisit soon.

PR #14, opened 2016-05-16
rv32ui: sh: Added side effect test

SebastianBoe:
From the test comment:
sh to a word aligned address should only affect the 2 lower bytes
and should leave the 2 upper bytes unmodified.

In this test we write 2 bytes to the lower 2 bytes of the word
tdat11 and then ensure that the both the upper 2 bytes and lower
2 bytes are as expected.

### DIFF
--- a/isa/rv32ui/sh.S
+++ b/isa/rv32ui/sh.S
@@ -68,6 +68,23 @@ RVTEST_CODE_BEGIN
   TEST_ST_SRC21_BYPASS( 22, 1, 1, lh, sh, 0x0011, 8, tdat );
   TEST_ST_SRC21_BYPASS( 23, 2, 0, lh, sh, 0x3001, 10, tdat );
 
+  #---------------------------------------------------------------
+  # Side effect tests
+  #---------------------------------------------------------------
+
+  # sh to a word aligned address should only affect the 2 lower bytes
+  # and should leave the 2 upper bytes unmodified.
+  #
+  # In this test we write 2 bytes to the lower 2 bytes of the word
+  # tdat11 and then ensure that the both the upper 2 bytes and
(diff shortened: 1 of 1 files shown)


SebastianBoe:
This test provokes an embarrassing bug I had where sh and sb wrote full 32bit words.
https://github.com/SebastianBoe/turborav/commit/1477ade270474831c7e219a6821133f981a42197

aswaterman:
this looks like a good thing to test -- I wonder if it should be generalized to the other tests, and if it can be written to avoid the TEST_CASE macro, which is perhaps too low-level.  I'm OK to merge either way.

SebastianBoe:
Hi,

I don't want to hurt the code maintainability so I made an attempt
 at using the higher level TEST_ST_OP macro. Unfortunately I found that 
it could not be used to provoke this failure.

Specifically
```Assembly
  TEST_ST_OP( 24, lw, sh, 0x12343098, 0, tdat11);

  generates

  TEST_CASE( 24, x3, 0x12343098, \
      la x1, tdat11;             \
      li x2, 0x12343098;         \
      sh x2, 0(x1);              \
      lw x3, 0(x1);              \
  )

  but we want

  TEST_CASE( 24, x3, 0x12343098, \
      la x1, tdat11;             \
      li x2, 0x00003098;         \
      sh x2, 0(x1);              \
      lw x3, 0(x1);              \
  )

tdat11: .word 0x12345678
```

aswaterman:
@colinschmidt is this compatible with the vector tests?

colinschmidt:
The vector tests already had to diverge from the standard load/store tests so this has no impact. Thanks for asking.

aswaterman:
ok, merging

PR #16, opened 2016-07-12
Add a "--with-xlen" configure argument

palmer-dabbelt:
The tools had riscv64-unknown-elf-gcc hard-coded all over the place.
This lets users override that if they have a 32-bit toolchain.


palmer-dabbelt:
@aswaterman thanks.

PR #24, opened 2016-08-23
Add requirements.txt and reorder imports by type.

richardxia:
@timsifive cc @mwachs5 

I added a requirements.txt with `pexpect` in it, and tested that it works with `pip install -r requirements.txt`.

I also went ahead and reordered the imports in `debug/gdbserver.py` and `debug/testlib.py` so that they are grouped into up to three categories, as recommended in [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports):

1. standard library imports
2. related third party imports
3. local application/library specific imports

This makes it clearer which imports come from the standard library and which imports need to be installed separately, such as `pexpect`.

I don't have a great way of testing this yet, since I'm still learning my way around these repos, so please let me know if this works for you still.


debug/testlib.py
```diff
@@ -1,17 +1,17 @@
...
 import time
 
+import pexpect
+
 # Note that gdb comes with its own testsuite. I was unable to figure out how to
 # run that testsuite against the spike simulator.
 
 def find_file(path):
-    for directory in (os.getcwd(), os.path.dirname(testlib.__file__)):
+    for directory in (os.getcwd(), os.path.dirname(__file__)):
```
richardxia: `__file__` is a special global variable in Python that is equivalent to `testlib.__file__`, but always for the current file in question. This removes the need to have to import itself.

debug/requirements.txt
```diff
@@ -0,0 +1 @@
+pexpect==4.2.1
```
timsifive: Is that exact version required? I'd prefer something like >=4 or whatever. Having to change this when we discover some other distro packages 4.2.0 or something seems like a lot of work.

richardxia: I just picked the latest version because I didn't know exactly which version was required. I've changed the requirement to >=4.0.0 in https://github.com/riscv/riscv-tests/pull/24/commits/de9053b96294acbcf2c2de46634720881f09f3a3, which I assume should be fairly safe.

PR #35, opened 2016-11-07
Tell OpenOCD to pick an unused port for gdb server

richardxia:
@timsifive @mwachs5 

We're still occasionally running into this spurious error in our internal CI system:

```pytb
Running SimpleS0Test ... Waiting for OpenOCD to examine RISCV core...
Traceback (most recent call last):
  File "rocket-chip/riscv-tools/riscv-tests/debug/gdbserver.py", line 710, in <module>
    sys.exit(main())
  File "rocket-chip/riscv-tools/riscv-tests/debug/gdbserver.py", line 702, in main
    return testlib.run_all_tests(module, target, parsed.test, parsed.fail_fast)
  File "rocket-chip/riscv-tools/riscv-tests/debug/testlib.py", line 287, in run_all_tests
    result = instance.run()
  File "rocket-chip/riscv-tools/riscv-tests/debug/testlib.py", line 384, in run
    self.classSetup()
  File "rocket-chip/riscv-tools/riscv-tests/debug/gdbserver.py", line 97, in classSetup
    self.gdb = gdb(self.target, self.server.port, self.binary)
  File "rocket-chip/riscv-tools/riscv-tests/debug/gdbserver.py", line 57, in gdb
    g.p("$priv=3")
  File "rocket-chip/riscv-tools/riscv-tests/debug/testlib.py", line 246, in p
    value = int(output.split('=')[-1].strip(), 0)
ValueError: invalid literal for int() with base 0: 'No registers.'
```

When inspecting `openocd.log`, you can see this at the bottom:

```
Error: 482 28849 server.c:265 add_service(): couldn't bind tcl to socket: Address already in use
```

This leads me to believe that the issue is with a race condition between `gdbserver.py` picking an unused port and OpenOCD binding to that port, whereby another process on the machine may bind to the port in between the two events.

This PR changes `testlib.py` to pass in the command `gdb_port 0` to OpenOCD, which tells OpenOCD to bind to a random unused port. One problem is that OpenOCD itself doesn't understand that making a syscall to bind to port 0 means that the OS will bind to a random port, so all of OpenOCD's logs just say that it's binding to port 0. In order to figure out which port was actually assign, I used `lsof` on the subprocess's open sockets and assert that there should be exactly one socket listening on a port.

Let me know if you can think of a better way of doing this, since it does feel a little bit messy.

Unrelated, but when I try to run the tests, they fail on `SimpleF18Test`, even when I'm on master. I tried making sure that I'm using the latest version of `riscv-isa-sim`, but that didn't seem to fix it.

### DIFF
--- a/debug/testlib.py
+++ b/debug/testlib.py
@@ -130,11 +130,10 @@ class Openocd(object):
         if debug:
             cmd.append("-d")
 
-        # Assign port
-        self.port = unused_port()
         # This command needs to come before any config scripts on the command
         # line, since they are executed in order.
-        cmd[1:1] = ["--command", "gdb_port %d" % self.port]
+        # Tell OpenOCD to bind to an unused port.
+        cmd[1:1] = ["--command", "gdb_port %d" % 0]
 
         logfile = open(Openocd.logname, "w")
         logfile.write("+ %s\n" % " ".join(cmd))
@@ -157,6 +156,31 @@ class Openocd(object):
             if not messaged and time.time() - start > 1:
                 messaged = True
                 print "Waiting for OpenOCD to examine RISCV core..."
+
+        self.port = self._get_gdb_server_port()
+
+    def _get_gdb_server_port(self):
+        """Get port that OpenOCD's gdb server is listening on."""
+        MAX_ATTEMPTS = 5
+        PORT_REGEX = re.compile(r'(?P<port>\d+) \(LISTEN\)')
+        for _ in range(MAX_ATTEMPTS):
+            with open(os.devnull, 'w') as devnull:
+                output = subprocess.check_output([
+                    'lsof',
+                    '-a',  # Take the AND of the following selectors
+                    '-p{}'.format(self.process.pid),  # Filter on PID
+                    '-iTCP',  # Filter only TCP sockets
+                ], stderr=devnull)
+            matches = list(PORT_REGEX.finditer(output))
+            if len(matches) > 1:
+                raise Exception(
+                    "OpenOCD listening on multiple ports. Cannot uniquely "
+                    "identify gdb server port.")
+            elif matches:
+                [match] = matches
+                return int(match.group('port'))
+            time.sleep(1)
+        raise Exception("Timed out waiting for gdb server to obtain port.")
 
     def __del__(self):
         try:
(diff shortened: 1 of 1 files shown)


debug/testlib.py
```diff
@@ -158,6 +157,31 @@ def __init__(self, cmd=None, config=None, debug=False):
...
+                ], stderr=devnull)
+            matches = list(PORT_REGEX.finditer(output))
+            if len(matches) > 1:
+                raise Exception(
+                    "OpenOCD listening on multiple ports. Cannot uniquely "
+                    "identify gdb server port.")
+            elif matches:
+                [match] = matches
+                return int(match.group('port'))
+            time.sleep(1)
```
timsifive: I'd reduce this to .1s or even less, and increase the number of attempts. 1s is an awful long time to wait when 10ms is probably enough.

debug/testlib.py
```diff
@@ -158,6 +157,31 @@ def __init__(self, cmd=None, config=None, debug=False):
...
+        PORT_REGEX = re.compile(r'(?P<port>\d+) \(LISTEN\)')
+        for _ in range(MAX_ATTEMPTS):
+            with open(os.devnull, 'w') as devnull:
+                output = subprocess.check_output([
+                    'lsof',
+                    '-a',  # Take the AND of the following selectors
+                    '-p{}'.format(self.process.pid),  # Filter on PID
+                    '-iTCP',  # Filter only TCP sockets
+                ], stderr=devnull)
+            matches = list(PORT_REGEX.finditer(output))
```
timsifive: list(regex.finditer()) is equivalent to regex.findall()

debug/testlib.py
richardxia: Sounds good, reduced the sleep time to 0.1s, and increased the max attempts to 50. Empirically, it looks like it took 6 sleeps (~600ms) for me for it to actually pick up the port, so it looks like it actually is a bit slow to start up the server, since I think it actually requires waiting long enough for the RTL simulator to connect to the JTAG interface.

debug/testlib.py
richardxia: They're actually slightly different. [`regex.findall()`](https://docs.python.org/2/library/re.html#re.findall) returns a list of strings or a list of tuples of strings if you have multiple capture groups. [`regex.finditer()`](https://docs.python.org/2/library/re.html#re.finditer) yields `MatchObjects`, which allow you to use named capture groups like I did.

timsifive: Interesting. I didn't realize that.

PR #41, opened 2017-02-10
debug: check that reading nonexistent CSR returns error.

mwachs5:
This isn't ready to merge, as it fails, and OpenOCD needs to be bumped.

@timsifive can you take a look? If I run this test with my new version of OpenOCD which returns errors, it does correctly print the expected string in the test log, but then the overall test fails. I think I'm not doing the actual `assertIn` correctly.

### DIFF
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -570,9 +570,18 @@ class WriteCsrs(RegsTest):
 class NonExistentCsr(RegsTest):
     def test(self):
         #Print old version of $misa at 0xF10 (moved to 0x301)
-        self.gdb.p("$csr3856")
-        #Expect this to return an error, not just a made-up value.
-        assertIn("Could not fetch register", output)
+        try:
+            self.gdb.p("$csr3856")
+            assert False, "Nonexistent CSR read should have failed."
+        except testlib.CannotAccessRegister as e:
+            assertEqual(e.regname, "csr3856")
+
+        #Try to write $misa in its old location
+        try:
+            self.gdb.p("$csr3856")
+            assert False, "Nonexistent CSR write should have failed."
+        except testlib.CannotAccessRegister as e:
+            assertEqual(e.regname, "csr3856")
 
 class DownloadTest(GdbTest):
     def setup(self):
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -567,6 +567,13 @@ class WriteCsrs(RegsTest):
         assertEqual(123, self.gdb.p("$x1"))
         assertEqual(123, self.gdb.p("$csr832"))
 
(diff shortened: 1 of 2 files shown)


debug/gdbserver.py
```diff
@@ -567,6 +567,13 @@ def test(self):
         assertEqual(123, self.gdb.p("$x1"))
         assertEqual(123, self.gdb.p("$csr832"))
 
+class NonExistentCsr(RegsTest):
+    def test(self):
+        #Print old version of $misa at 0xF10 (moved to 0x301)
+        self.gdb.p("$csr3856")
```
timsifive: output = ...

mwachs5: Sorry, not getting it. I tried 
`output = self.gdb.p("..")` 
but that's wrong.  Maybe it's on stderr or something:

```
(gdb) p/x $csr3856
p/x $csr3856
Could not fetch register "csr3856"; remote failure reply 'E0E'
(gdb) ////////////////////////////////////////
::::::::[ ran 2 tests in 124s ]:::::::::
1 tests returned exception
    NonExistentCsr
1 tests returned pass
```

mwachs5: NM I read testlib.py and get why it doesn't work. Will figure it out.

timsifive: Your test returned an exception. There should be more output that details exactly what happened, but I suspect you hit this code in p():
```
    def p(self, obj):
        output = self.command("p/x %s" % obj)
        m = re.search("Cannot access memory at address (0x[0-9a-f]+)", output)
        if m:
            raise CannotAccess(int(m.group(1), 0))
        value = int(output.split('=')[-1].strip(), 0)
        return value
```

So your tests needs to:
```
try:
   self.gdb.p(...)
   assert False, "access should have failed"
except CannotAccess:
   # expected
   pass
```

just like MemTestReadInvalid

mwachs5:
Ok, this works now for this test, but for the 64-bit tests I now get an error about -march needing to be set to use 'D' instructions. I hadn't bumped my toolchain in a while, does something need to change in the way I'm invoking the test?

mwachs5:
Actually this doesn't work against Spike  yet, still digging in.

mwachs5:
It works in Spike if I just run Spike and GDB, but I get a timeout if I try to run it through GDB Server...

mwachs5:
@timsifive what does it mean to "pass" these tests? Can I just run make debug-check? Or something else?

timsifive:
`make debug-check` should do it. What I do is I cd into the debug
directory, and run `make clean && make all`. This runs tests against both
32- and 64-bit spike. Testing against hardware you have to manually by
running gdbserver.py directly.

PR #44, opened 2017-03-30
Two GP-related fixes

palmer-dabbelt:
The second one is a preemptive fix, the first one effectively turns GP back on (ld won't relax to it unless it's there, and the symbol name changed).

### DIFF
--- a/benchmarks/common/crt.S
+++ b/benchmarks/common/crt.S
@@ -108,7 +108,10 @@ _start:
 1:
 
   # initialize global pointer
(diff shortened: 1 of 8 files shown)


aswaterman:
I've cherry-picked these onto the priv-1.10 branch https://github.com/riscv/riscv-tests/tree/priv-1.10 which I intend to merge into master in a few days, so should we just close this?

palmer-dabbelt:
WFM

PR #47, opened 2017-05-15
Debug 0.13 Tests

mwachs5:
What else is required to get the test changes merged in...? Do we need to add them back to the test regression?


mwachs5:
Which ones? I think the "*Invalid" ones don't really work, are there others?

palmer-dabbelt:
@mwachs5 The tests should all pass on Spike now, I think we should merge this.

mwachs5:
sounds good to me

mwachs5:
Does 'debug-check' actually run?

palmer-dabbelt:
With my latest commit, "debug-check" runs.  We need to merge spike before we can merge this.

PR #48, opened 2017-05-15
Get the test running on Spike again

palmer-dabbelt:


debug/Makefile
```diff
@@ -29,7 +12,7 @@ pylint:
 	pylint --rcfile=pylint.rc *.py
 
 %.log:
-	$(GDBSERVER_PY) --isolate --$(subst .log,,$@) --server_cmd $(OPENOCD_DIR)/bin/openocd \
+	$(GDBSERVER_PY) --isolate --$(subst .log,,$@) --server_cmd openocd \
```
mwachs5: should this be $(RISCV)/bin/openocd ?

debug/gdbserver.py
```diff
@@ -612,69 +615,6 @@ def test(self):
         assertEqual(self.gdb.p("status"), self.crc)
         os.unlink(self.download_c.name)
 
-class MprvTest(GdbTest):
```
mwachs5: Why are these disabled?

debug/Makefile
```diff
@@ -12,7 +12,11 @@ pylint:
 	pylint --rcfile=pylint.rc *.py
 
 %.log:
-	$(GDBSERVER_PY) --isolate --$(subst .log,,$@) --sim_cmd $(RISCV)/bin/$(RISCV_SIM) \
+	$(GDBSERVER_PY) \
+		--isolate \
+		--$(subst .log,,$@) \
+		--sim_cmd $(RISCV)/bin/$(RISCV_SIM) \
+		--server_cmd openocd \
```
mwachs5: shouldn't this be $(RISCV)/bin/openocd?

debug/gdbserver.py
palmer-dabbelt: Isn't *PRIV no longer a register?  It's no longer recognized by binutils, and it had been emulated by OpenOCD for a while to write some other CSRs.

debug/Makefile
palmer-dabbelt: Thanks, fixed.

debug/gdbserver.py
mwachs5: It has always been like that. It's in the spec as a "virtual register"

palmer-dabbelt:
OK, so I just commented out the PRIV tests instead of deleting them.  I'll have to read the spec closer... :)

PR #53, opened 2017-05-23
Fail if simulator exits early.

richardxia:
Currently, if VCS has an issue and crashes before setting up the JTAG VPI port, then `testlib.py` will hang indefinitely looking for the port number in the output. This adds a check to the port number polling loop to raise an exception if the process has exited before we see the port number.


richardxia:
@palmer-dabbelt @mwachs5

palmer-dabbelt:
Thanks!

PR #54, opened 2017-05-26
Enable the RTOS when running tests on Spike

palmer-dabbelt:
You'll need my latest OpenOCD for this to actually work.


palmer-dabbelt:
The tools PR https://github.com/riscv/riscv-tools/pull/85 points to 4660402, so if this gets rebased that needs to be updated.

PR #55, opened 2017-06-23
Test gdb/OpenOCD during regular test run.

timsifive:


timsifive:
Passes in my setup. I don't know how to have Travis test something without merging it. (Even more complex here since the actual invocation is in a different repo.)

PR #56, opened 2017-06-26
Move target definition into individual files.

timsifive:
Instead of defining each target in targets.py, now each target gets its
own .py file. This means people can easily keep their own target files
around that they may not want to put into the main test source. As part
of that, I removed the freedom-u500-sim target since I assume it's only
used internally at SiFive.

Added a few cleanups as well:
* Update README examples, mostly --sim_cmd instead of --cmd.
* Allow defining misa in a target, to skip running of ExamineTarget.
* Rename target.target() to target.create(), which is less confusing.
* Default --sim_cmd to `spike`
* Got rid of `use_fpu`, instead looking at F or D in $misa.


debug/README.md
```diff
@@ -12,37 +12,47 @@ Targets
...
+`./gdbserver.py targets/RISC-V/spike32.py`
 
-32-bit SiFive Core on Supported FPGA Boards &  Hardware
--------------------------------------
+32-bit SiFive Core on Supported FPGA Boards & Hardware
+------------------------------------------------------
 
-`./gdbserver.py --freedom-e300`
-`./gdbserver.py --hifive1`
+`./gdbserver.py targets/SiFive/E300.py`
```
mwachs5: Is this specifically something like "E300 Olimex Dongle"?

timsifive: Yes.
I remove the simulation target because I thought it was just a SiFive internal thing. If not, I can add a config file for it.

PR #58, opened 2017-07-06
debug: Make the 'out of reset' tests apply reset

mwachs5:
I'm not sure what the intention is for the "Instant" tests. To make them work on real hardware, doesn't reset need to be applied? With this patch the InstantHaltTest works on FPGA images for me. Were they already working on real hardware for you @timsifive?


timsifive:
I've never had InstantHaltTest working on FPGA/hardware. You're right that a reset needs to be applied to make it work.

PR #62, opened 2017-08-08
rv64[ms]i-csr: Only emit F instructions when compiled for F.

richardxia:
Allow this test to be compiled when `-march` lacks the `f` flag by checking for the `__riscv_flen` compiler macro. The `else` case unconditionally fails so that the test errors out if compiled without `f` and run on a core that supports `f`.


richardxia:
cc @palmer-dabbelt @aswaterman

Issue #65, opened 2017-08-14
debug: OpenOCD examine timeout can't be hard-coded


mwachs5:
Currently the timeout for OpenOCD to report that it is ready for  GDB connections is hard-coded to 60 seconds. This should be a parameter. For a simulation target this is way too short.

timsifive:
Can this be closed since #66 merged?

mwachs5:
Closed by #66

PR #66, opened 2017-08-14
Debug: Usability Features

mwachs5:
* Print out the path the the temporary OpenOCD/ GDB log files, as for long running simulations it's nice to see what these are doing
* Allow the timeout for OpenOCD startup to be configurable per-target, as simulation targets may take a long time. Addresses #65

### DIFF
--- a/debug/testlib.py
+++ b/debug/testlib.py
@@ -561,8 +561,6 @@ class BaseTest(object):
                 result = "fail"
             else:
                 result = "exception"
-                header ("Backtrace")
-                print e
             if isinstance(e, TestFailed):
                 header("Message")
                 print e.message
--- a/debug/testlib.py
+++ b/debug/testlib.py
@@ -282,7 +282,7 @@ class Gdb(object):
     def __init__(self,
             cmd=os.path.expandvars("$RISCV/bin/riscv64-unknown-elf-gdb")):
         self.child = pexpect.spawn(cmd)
-        self.child.logfile.write("+ %s\n" % cmd)
+        Gdb.logfile.write("+ %s\n" % cmd)
         self.wait()
         self.command("set confirm off")
         self.command("set width 0")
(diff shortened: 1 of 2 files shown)


debug/testlib.py
```diff
@@ -161,8 +161,9 @@ def __del__(self):
 class Openocd(object):
     logfile = tempfile.NamedTemporaryFile(prefix='openocd', suffix='.log')
     logname = logfile.name
+    print "OpenOCD Temp Logfile: %s " % logname
```
timsifive: Why the trailing space?

debug/testlib.py
```diff
@@ -276,11 +277,11 @@ def __init__(self, address):
 class Gdb(object):
     logfile = tempfile.NamedTemporaryFile(prefix="gdb", suffix=".log")
     logname = logfile.name
+    print "GDB Temporary file: %s" % logname
```
timsifive: I'd call these
"OpenOCD logfile" and "GDB logfile". At least be consistent in capitalization and temp/temporary, logfile/file

mwachs5: I think we should call them temporary, as my understanding is that these get deleted when the program exits, right?

mwachs5:
Personally I like the logfile format vs printing to stdout

debug/testlib.py
timsifive: That's right. I'm happy to call them temporary, as long as it's consistent in both print statements.

mwachs5: Sure, will clean up.

timsifive:
Now in the log file nothing shows up in the gdb section, eg. on a passing test:
```
----------------------------[ /tmp/gdbflnU_f.log ]----------------------------

-------------------------------[ End of logs ]--------------------------------
```

mwachs5:
Oops, sorry. I never use the GDB log I guess. If I didn't make the above change, I got an error about NoneType (self.child.logfile)

PR #69, opened 2017-08-28
Proper multicore support for debug tests

timsifive:
Note that this changes the format of the .py files used to define targets. Now targets contain a list of (potentially heterogeneous) harts, whereas before the number was autodetected and assumed to be homogeneous.

Link files now must define __bss_start and __bss_end. The memory between those two will be zeroed in entry.S.

Individual tests can decide whether they are written to just test one hart, or to be aware of multiple harts. In the former case they are automatically given a random hart to run against, and the harts not tested are automatically parked in a loop in 

Also:
* Make MemTestBlock error message more readable.
* A new MulticoreRunHaltStepiTest does some sanity checking on run/halt/stepi in multicore.


timsifive:
This branch should be the path forward in debug testing. Can one of you please look at it?

PR #76, opened 2017-09-18
Add interrupts to MulticoreRunHaltStepiTest.

timsifive:
Just to hammer on anything at once, and hopefully catch weird
interactions if they exist.

### DIFF
--- a/debug/programs/multicore.c
+++ b/debug/programs/multicore.c
@@ -1,5 +1,7 @@
 #include <stdint.h>
 
+#include "init.h"
+
 typedef struct {
     int counter;
 } atomic_t;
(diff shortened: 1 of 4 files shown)


debug/programs/multicore.c
```diff
@@ -44,19 +38,31 @@ void put_lock(atomic_t *lock)
...
 
 int main()
 {
     uint32_t hartid = csr_read(mhartid);
     hart_count[hartid] = 0;
+    interrupt_count[hartid] = 0;
+
+    set_trap_handler(increment_count);
+    // Despite being memory-mapped, there appears to be one mtimecmp
+    // register per hart. The spec does not address this.
```
palmer-dabbelt: @aswaterman This isn't _explicitly_ mentioned in the spec, but since these are in the CSR where everything else is per-hart I don't know if it's explicitly worth calling this out.

PR #79, opened 2017-09-29
Multigdb support

timsifive:
The Gdb class now can handle connecting to more than one gdb. It
    enumerates the harts across all connections, and when asked to select a
    hart, it transparently sends future gdb commands to the correct
    instance.
    
    Multicore tests still have to be aware of some differences. The main one
    is that when executing 'c' in RTOS mode, all harts resume, while in
    multi-gdb mode only the current one resumes. Additionally, gdb doesn't
    set breakpoints until 'c' is issued, so the hart where breakpoints are
    set needs to be resumed before other harts might see them.


timsifive:
If you have a usecase for the tests that's important to you and not spike/HiFive1/random FPGA image I have, then please take some time to make sure it still works.

Note that triggers won't work right in multi-gdb mode until you first approve https://github.com/riscv/riscv-openocd/pull/116.

timsifive:
Are you going to take the time to test this, or should I just merge it?

mwachs5:
go for it

PR #92, opened 2017-11-01
SBREAK test now checks EPC value.

### LINKED ISSUE #89: SBREAK test does not check EPC
EBREAK and ECALL set EPC register to the address of the EBREAK/ECALL instruction, per the Privileged Specification v1.11. 

The `rv64si/sbreak.S` test only checks the `cause` register, not the `epc` register.

Looks like I've been called out: https://github.com/ucb-bar/riscv-boom/issues/39

ccelio:
Closes #89


aswaterman:
thanks

Issue #93, opened 2017-11-01
debug: `assert threads` failure


mwachs5:
I am running riscv-tests against a simulator, and get an assertion failure after GDB attempts to do `info threads`. If I run everything manually, everything works fine, but I am suspicious that it may be attempting to do `info threads` before GDB has finished connecting to OpenOCD. However, with this error, no GDB log is generated so it is hard to debug.

What timeout/handshake is used to ensure that `info threads` is not called before GDB finishes connecting to OpenOCD?

Also the code which printed the temporary GDB file paths to the stdout has been removed, which makes it impossible to monitor the progress of the tests on a simulation.

mwachs5:
The issue is a timeout one. The `set remotetimeout` command should be called BEFORE the `target extended remote localhost` command. This is not being done currently. I will have a PR with this change.

https://github.com/riscv/riscv-tests/blob/master/debug/testlib.py#L343

This addresses the thing I was trying to debug, but not the issue that it's very difficult to debug without the GDB output, which is not getting created in this failure mode even though GDB has already started running.

mwachs5:
I am closing this as it was fixed by #94

timsifive:
> with this error, no GDB log is generated so it is hard to debug.

Fixed in d2bf7e8007f2f8972f3974ab985c26cba38c18ec

PR #94, opened 2017-11-01
debug: Need to apply remotetimeout before connecting to remote target

mwachs5:
Addresses #93 
The `target extended-remote localhost` command itself is a long-running command in need of a timeout, so the `set remotetimeout` command needs to be called before actually connecting to the target, not after.

### DIFF
--- a/debug/testlib.py
+++ b/debug/testlib.py
@@ -504,8 +504,8 @@ class Gdb(object):
             if m:
                 threads.append(Thread(*m.groups()))
         assert threads
-        # >>>if not threads:
-        # >>>    threads.append(Thread('1', '1', 'Default', '???'))
+        #>>>if not threads:
+        #>>>    threads.append(Thread('1', '1', 'Default', '???'))
         return threads
 
     def thread(self, thread):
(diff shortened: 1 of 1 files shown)


mwachs5:
well, it may just be re-masking some other issue, but it does resolve my problem when connecting to simulations

debug/testlib.py
```diff
@@ -748,16 +750,15 @@ def classSetup(self):
         BaseTest.classSetup(self)
 
         if gdb_cmd:
-            self.gdb = Gdb(self.server.gdb_ports, gdb_cmd, binary=self.binary)
+            self.gdb = Gdb(self.server.gdb_ports, gdb_cmd, timeout = self.target.timeout_sec,  binary=self.binary)
         else:
-            self.gdb = Gdb(self.server.gdb_ports, binary=self.binary)
+            self.gdb = Gdb(self.server.gdb_ports, timeout = self.target.timeout_sec, binary=self.binary)
```
richardxia: This will probably fail the linter (I *think*; I haven't used pylint myself, but I assume it enforces PEP 8). Python's official style guide is to not put any spaces around the equals sign in named arguments. Also only one space after commas is allowed.

Issue #96, opened 2017-11-03
debug: --32 and --64 arguments seem ignored


mwachs5:
IN gdbserver.py, there are arguments for the command line, `--32` and `--64`, which are supposed to allow knowing the XLEN, used for the GDB `set arch` command. But all they do is set target.xlen, but it is hart.xlen which is used for everything. In addition, `$misa` is read anyway and XLEN is determined from that.

This doesn't show up in the tests under regression because those have separate Hart subclasses which all specify their xlen explicitly.

So what is the point of the command line argument? What is the expected relationship between the command line argument, the `xlen` value that the hart subclasses specify, and the `$misa` value determined during ExamineTarget?

mwachs5:
Also, if you are relying on the "reading MISA" thing to even print out XLEN if you've guessed wrong about what XLEN is, you need to change the masking code, because you are likely reading a 64-bit value and there is no masking to make the read misa value be only 32 bits before the comparison here:

https://github.com/riscv/riscv-tests/blob/master/debug/testlib.py#L828

All of this contributes to this sort of output if you specify --32 on the command line without specifying xlen in the hart:

```
(gdb) target extended-remote localhost:40987
target extended-remote localhost:40987
Remote debugging using localhost:40987
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
0x00007fff00010050 in ?? ()
(gdb) info threads
info threads
  Id   Target Id         Frame 
* 1    Thread 1 (Name: Hart 0, RV32) 0x00007fff00010050 in ?? ()
(gdb) set arch riscv:rv0
set arch riscv:rv0
Undefined item: "riscv:rv0".
```

PR #97, opened 2017-11-03
Debug: Use the --32 and --64 command line arguments

mwachs5:
This is to address #96, but I'm not sure this is the change that makes sense.

### DIFF
--- a/debug/targets.py
+++ b/debug/targets.py
@@ -180,5 +180,10 @@ def target(parsed):
 
     t = found[0](parsed.target, parsed)
     assert t.harts, "%s doesn't have any harts defined!" % t.name
+    for h in t.harts :
+        if (h.xlen == 0):
+            h.xlen = parsed.xlen
+        elif (h.xlen != parsed.xlen):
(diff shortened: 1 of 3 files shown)


debug/targets.py
```diff
@@ -180,5 +180,10 @@ def target(parsed):
 
     t = found[0](parsed.target, parsed)
     assert t.harts, "%s doesn't have any harts defined!" % t.name
+    for h in t.harts :
+        if (h.xlen == 0):
+            h.xlen = parsed.xlen
+        elif (h.xlen != parsed.xlen):
+            raise Exception("Hart had a specified XLEN of %d and target had a specified XLEN of %d, they must match." % (h.xlen, parsed.xlen))
```
timsifive: Originally this option would allow you to override XLEN as well. Given that changing XLEN often requires a new link map I don't know how useful that really was, so I'm fine with this new requirement.

I'd say this instead:
"The target has an XLEN of %d, but the command line specified an XLEN of %d. They must match."

PR #107, opened 2017-11-22
Check sepc for rv64si/scall test.

### LINKED ISSUE #105: rv64si/scall does not check epc.
I'm finding the logic a little hard to follow, but it looks like the `scall` test only checks `scause` but not `sepc`.

ccelio:
Closes #105.


aswaterman:
thanks!

PR #109, opened 2017-11-30
Clean up VcsSim init()

timsifive:
Use a unique log file, so you can run multiple instances at once.
Add time out to waiting for the simulator to be ready.

### DIFF
--- a/debug/testlib.py
+++ b/debug/testlib.py
@@ -142,9 +142,10 @@ class Spike(object):
         return self.process.wait(*args, **kwargs)
 
 class VcsSim(object):
-    logname = "simv.log"
+    logfile = tempfile.NamedTemporaryFile(prefix='simv', suffix='.log')
+    logname = logfile.name
 
-    def __init__(self, sim_cmd=None, debug=False):
+    def __init__(self, sim_cmd=None, debug=False, timeout=300):
         if sim_cmd:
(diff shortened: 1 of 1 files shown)


debug/testlib.py
```diff
@@ -142,9 +142,10 @@ def wait(self, *args, **kwargs):
         return self.process.wait(*args, **kwargs)
 
 class VcsSim(object):
-    logname = "simv.log"
+    logfile = tempfile.NamedTemporaryFile(prefix='simv', suffix='.log')
+    logname = logfile.name
 
-    def __init__(self, sim_cmd=None, debug=False):
+    def __init__(self, sim_cmd=None, debug=False, timeout=300):
```
mwachs5: is there a path to override this timeout? Let's not hard-code any timeouts.

mwachs5: For example can target class have a "init_timeout" variable which we pass in here but can be overridden in the target subclass?

timsifive: Yes. The function is called from the .py file containing a target configuration, so each target can specify its own timeout.

timsifive:
The timeout could be configured eg. at https://github.com/riscv/riscv-tests/blob/master/debug/targets/SiFive/Freedom/E300Sim.py#L17

PR #119, opened 2018-03-19
Update Makefile to allow for RISCV_PREFIX to be set by the configure 

rishikhan:
Update Makefile to allow for RISCV_PREFIX to be set by the configure --target


rishikhan:
The reason for this change is to allow compilers other than riscv64-unknown-elf-gcc to work. If you have a different vendor or OS, they can be passed to configure and everything works as planned.

Issue #122, opened 2018-04-09
debug: propagate GDB command timeout more places?


mwachs5:
I'm not sure how the GDB command timeout is passed through. For example, the `info threads` command can do quite a bit and take quite a long time on a slow (e.g. simulation), multi-core target:

https://github.com/riscv/riscv-tests/blob/45380af7d42ee3302fc229030694f8ea4506d79f/debug/testlib.py#L562

Is the target-specified timeout passed through at all in this case? If not, can we pass it through *every* time we call `gdb.command(...)`?

timsifive:
timeout_sec is just something we tell gdb for its internal timeout. It's the maximum time gdb waits to hear back from the gdbserver (openocd in our case) before erroring out.

There's a separate timeout imposed by our test code, which is not configurable. It doesn't make sense for that timeout to be less than what we tell gdb, though, so I'll make a change to use the same configurable value for that timeout.

timsifive:
Do you have a log or some detail on which timeout you ran into? I want to make sure I cover that code path.

mwachs5:
#123 seems to do a good job here. Thanks! May need to tweak in the future but overall approach seems good.

PR #123, opened 2018-04-10
Compute gdb command timeout based on ops estimate

timsifive:
The caller of gdb.command() should estimate how much work gdb needs to
do, and testlib then scales this up proportional to the general gdb
timeout we configured. This hopefully allows us to configure a tighter
timeout, so we don't have to have a multi-hour timeout just for
something that takes long like `load` on a really slow simulator.

Hopefully this addresses #122.

This passed a rocket-chip build in https://github.com/freechipsproject/rocket-chip/pull/1327


mwachs5:
Thanks for this. I am trying it out now on my slow target.

Issue #131, opened 2018-04-27
debug: need to ensure $satp is 0 before attemping to switch priv modes


mwachs5:
The priv tests need to ensure that $satp is 0 before attempting to switch priv modes, similar to how they have to set up the PMP bits before switching priv modes, otherwise you'll get an access exception.

PR #132, opened 2018-04-27
debug: need to clear satp before changing priv

mwachs5:
ISA Manual does not require this register to be reset, and attempting to execute code with VM enabled when VM hasn't been set up is going to just lead to sadness 50% of the time.


mwachs5:
fixes #131

PR #135, opened 2018-04-30
[rv64ua/lrsc] Initialize memory read out.

### LINKED ISSUE #134: rv64ua/lrsc.S reads from un-initialized memory.
**Background:**

The `lrsc` test performs a `load-reserve` from un-initialized memory. The write back value is `dontCare`, however, this can be problematic for co-simulation.

https://github.com/riscv/riscv-tests/blob/835f252ebe59ef782e96f467462a03754ea87936/isa/rv64ua/lrsc.S#L34-L41

It loads the address of `foo`, but then adds 1024 to it, in the hopes that 1024 is beyond the reach of the reservation (a painfully implementation-dependent value).

--------
**Question:**

Is leaving `foo+1024` un-initialized an oversight, or a deliberate choice? Is it wanting a smaller static binary to hasten loading?

ccelio:
Even though the load contents are discarded, this un-initialized memory value
can lead to a divergence for co-simulation between two different RISC-V designs.

Closes #134.

### DIFF
--- a/isa/rv64ua/lrsc.S
+++ b/isa/rv64ua/lrsc.S
@@ -35,7 +35,7 @@ TEST_CASE( 2, a4, 1, \
 # TODO is this actually mandatory behavior?
 TEST_CASE( 3, a4, 1, \
   la a0, foo; \
-  add a1, a0, 1024; \
+  la a1, fooTest3; \
   lr.w a1, (a1); \
   sc.w a4, a1, (a0); \
 )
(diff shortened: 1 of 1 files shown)


ccelio:
This is the resulting disassembly:

```
Disassembly of section .data:            
                                         
0000000080002000 <begin_signature>:      
    80002000:  0000                 unimp
    80002002:  0000                 unimp
                                         
0000000080002004 <barrier>:              
    80002004:  0000                 unimp
    80002006:  0000                 unimp
                                         
0000000080002008 <foo>:                  
    80002008:  0000                 unimp
    8000200a:  0000                 unimp
    8000200c:  0000                 unimp
...
0000000080002400 <fooTest3>:             
    80002400:  0000                 unimp
    80002402:  0000                 unimp
    80002404:  0000                 unimp
    80002406:  0000                 unimp
```

PR #141, opened 2018-05-19
Fix MulticoreRunHaltStepiTest

timsifive:
The test actually wasn't checking interrupt counts at all. Fixing it
required some other changes:
Make sure all harts get to run
Add some retries, since on a loaded machine against spike both harts
might not get to run, even if you give spike a generous amount of time
to do so.

### DIFF
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -562,30 +562,48 @@ class MulticoreRunHaltStepiTest(GdbTest):
         self.gdb.load()
         for hart in self.target.harts:
             self.gdb.select_hart(hart)
+            self.gdb.p("$mhartid")
             self.gdb.p("$pc=_start")
 
     def test(self):
         previous_hart_count = [0 for h in self.target.harts]
         previous_interrupt_count = [0 for h in self.target.harts]
-        for _ in range(10):
-            self.gdb.c(wait=False)
-            time.sleep(2)
-            self.gdb.interrupt()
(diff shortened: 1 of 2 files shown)


debug/gdbserver.py
```diff
@@ -562,30 +562,48 @@ def setup(self):
...
+                    self.gdb.select_hart(h)
+                    pc = self.gdb.p("$pc")
+                    self.gdb.stepi()
+                    stepped_pc = self.gdb.p("$pc")
+                    assertNotEqual(pc, stepped_pc)
+                previous_hart_count = hart_count
+                previous_interrupt_count = interrupt_count
+                if ok:
+                    break
+            else:
```
mwachs5: Reading on phone so spacing may be off but what if does this else go with?

timsifive: It goes with the inner for (which does the retries if necessary).
Python allows an else clause on for/while loops, which get executed if the loop exits normally (eg. not due to break).

timsifive:
I have not reviewed any other tests. The reason this was false-passing is specific to this test.

debug/gdbserver.py
mwachs5: TIL! Thanks.

mwachs5:
I'm happy if this passes on Non-RTOS multicore Spike. We should probably review the other tests for false-passing, but doesn't need to block this PR.

PR #151, opened 2018-08-18
Fix to solve the failing tests shamt, csr and scall

srivatsa611y:


isa/rv64si/csr.S
```diff
@@ -81,6 +81,22 @@ RVTEST_CODE_BEGIN
...
+  li t0, 0x07   # giving read, write and execute permissions
+  or t0, t0, 8  # setting mode to TOR
+  li t1, 255
+  csrrc t1, pmpcfg0, t1
+  csrrs t1, pmpcfg0, t0
+  csrr  t1, pmpcfg0
+  andi  t1, t1, 255
+  bne t0, t1, fail
+  .global no_pmp_implemented
+  no_pmp_implemented:
```
aswaterman: Shouldn't this all be guarded by #ifdef __MACHINE_MODE (i.e., should the preceding #endif be moved down here)?  This code shouldn't execute in the S-mode variant of this test.

isa/rv64si/csr.S
```diff
@@ -81,6 +81,22 @@ RVTEST_CODE_BEGIN
   andi a0, a0, 1  # a0 = a0 & 1
   beqz a0, finish # if no user mode, skip the rest of these checks
 #endif /* __MACHINE_MODE */
+  la t0, user_mode_end
+  srli t0, t0, 2
```
aswaterman: To improve the readability of this code, please use the PMP_SHIFT macro, instead of the magic constant 2.

You can see the PMP macros here: https://github.com/riscv/riscv-test-env/blob/master/encoding.h#L135

isa/rv64si/csr.S
```diff
@@ -81,6 +81,22 @@ RVTEST_CODE_BEGIN
...
+  csrr t1, pmpcfg0
+  csrw pmpaddr0, t0
+  csrr t1, pmpaddr0
+  bne t0, t1, fail
+  li t0, 0x07   # giving read, write and execute permissions
+  or t0, t0, 8  # setting mode to TOR
```
aswaterman: Similarly, PMP_R | PMP_W | PMP_X | PMP_TOR instead of 0x7 | 8.

isa/rv64si/scall.S
```diff
@@ -26,16 +28,33 @@ RVTEST_CODE_BEGIN
...
   # Otherwise, if in S mode, then U mode must exist and we don't need to check.
+
+  la t0, user_mode_end
+  srli t0, t0, 2
+  csrr t1, pmpcfg0
+  csrw pmpaddr0, t0
+  csrr t1, pmpaddr0
+  bne t0, t1, fail
+  li t0, 0x07   # giving read, write and execute permissions
+  or t0, t0, 8  # setting mode to TOR
```
aswaterman: Same comment about PMP_ macros applies here

isa/rv64si/scall.S
```diff
@@ -26,16 +28,33 @@ RVTEST_CODE_BEGIN
...
 
-  # This is the expected trap code.
-  li t1, CAUSE_USER_ECALL
-
 #ifdef __MACHINE_MODE
   # If running in M mode, use mstatus.MPP to check existence of U mode.
   # Otherwise, if in S mode, then U mode must exist and we don't need to check.
+
+  la t0, user_mode_end
+  srli t0, t0, 2
```
aswaterman: Same comment about PMP_SHIFT macro applies here.

aswaterman:
Thanks, LGTM after these issues are addressed.

isa/rv64si/csr.S
srivatsa611y: Yes, you are correct

PR #152, opened 2018-08-21
Changing the register mstatus is read to

srivatsa611y:
The mstatus reading overwrites the expected user mode cause value. this in turn causes the test to fail when a config has user mode


aswaterman:
Ah, nice catch.

PR #161, opened 2018-09-08
RV64 s{ll,ra,rl}w tests with non-canonical values

tommythorn:
The xxW variants should also be tested with values that are _not_ canonical 32-bit in 64-bit values.
As best I can tell, these pass of the current version of Spike.


aswaterman:
Thanks.

PR #165, opened 2018-10-08
Tweak debug tests to run out of flash.

timsifive:
Not all tests pass when run out of flash yet, but it's getting a lot
closer. The ones still failing on HiFive1-flash are: DebugSymbols,
Hwbp2, InstantHaltTest, TriggerDmode, TriggerLoadAddressInstant, and
TriggerStoreAddressInstant.

### DIFF
--- a/debug/programs/debug.c
+++ b/debug/programs/debug.c
@@ -53,11 +53,8 @@ int main()
 
     volatile int i = 0;
     int j = 0;
-    static char fox[43];
+    char fox[] = "The quick brown fox jumps of the lazy dog.";
     unsigned int checksum = 0;
-
-    /* fox needs to be writable, but the string could be in ROM. */
-    strcpy(fox, "The quick brown fox jumps of the lazy dog.");
 
 start:
     while (i)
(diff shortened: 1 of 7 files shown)


timsifive:
Now all tests pass from flash (except for InstantHaltTest, which doesn't pass on HiFive1 at all).

debug/programs/debug.c
```diff
@@ -53,9 +53,12 @@ int main()
 
     volatile int i = 0;
     int j = 0;
-    char *fox = "The quick brown fox jumps of the lazy dog.";
+    static char fox[43];
     unsigned int checksum = 0;
 
+    /* fox needs to be writable, but the string could be in ROM. */
+    strcpy(fox, "The quick brown fox jumps of the lazy dog.");
+
```
aswaterman: If you revert to the original code, but change `char *fox = "...";` to `char fox[] = "...";`, it'll accomplish what you want without the need for a static variable or strcpy.

timsifive: That's a subtlety I hadn't considered. Thanks.

PR #169, opened 2018-11-13
Added MemTestBlockReadInvalid verifying the corresponding OpenOCD fix

cgsfv:
Testing that memory reads with OpenOCD through the DTM Version 0.13 does not fail even if the memory range generates an exception from target, but instead returns zero values for the memory locations that is not really readable.

### DIFF
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -237,7 +237,7 @@ class MemTestBlockReadInvalid(GdbTest):
     real_values = "EF BE AD DE 78 56 34 12"
 
     def early_applicable(self):
-        return self.target.uses_dtm_version_013
+        return self.target.invalid_memory_returns_zero
 
     def test(self):
         self.gdb.p("*((int*)0x%x) = 0xdeadbeef" % (self.hart.ram + 0))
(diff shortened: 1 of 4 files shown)


debug/gdbserver.py
```diff
@@ -232,6 +232,58 @@ def test(self):
...
+                         self.hart.ram + self.hart.ram_size + 8,
+                         self.zero_values)
+
+    def memory_test(self, start_addr, end_addr, expected_values):
+        dump = tempfile.NamedTemporaryFile(suffix=".simdata")
+        self.gdb.command("dump verilog memory %s 0x%x 0x%x" % (dump.name, start_addr, end_addr))
+        self.gdb.command("shell cat %s" % dump.name)
+        line = dump.readline()
+        line = dump.readline()
+        assertEqual(line.strip(' \t\n\r'), expected_values)
```
timsifive: You can just use line.strip(). By default strip strips all whitespace.

debug/targets.py
```diff
@@ -79,6 +79,9 @@ class Target(object):
     # hardware will every do that.
     implements_custom_test = False
 
+    # Target uses the DTM Version 0.13 for memory accesses
+    uses_dtm_version_013 = False
```
timsifive: I don't think this is what you want. The debug spec doesn't say what should happen when reading invalid memory, and in any case the DTM is completely memory-agnostic. The DM (through abstract commands) deals with memory, but in our case it's mainly the hart implementation that determines the behavior. I would call this something like invalid_memory_returns_zero, with a comment saying that when true reading invalid memory doesn't return an error.

debug/gdbserver.py
cgsfv: Agree.

debug/targets.py
cgsfv: Agree that this is more descriptive of what the flag actually means.

PR #171, opened 2018-11-16
Test memory content on failing SC

zarubaf:
Small test that will check that a failing store conditional will not commit into memory.


zarubaf:
That test would have saved me three weeks of debugging as torture is not emitting any lr/sc pairs and they only show up very late in the Linux boot process.

Just thought this might save somebody some time in the future.

PR #174, opened 2018-12-31
Fix MemTestBlock

timsifive:
This test used to false pass on 64-bit targets because gdb doesn't like
using Intel hex files on 64-bit targets.

### DIFF
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -70,6 +70,8 @@ def srec_parse(line):
     elif typ == 'S7':
         # ignore execution start field
         return 7, 0, 0
+    else:
+        raise TestFailed("Unsupported SREC type %r." % typ)
 
 def readable_binary_string(s):
     return "".join("%02x" % ord(c) for c in s)
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -51,16 +51,25 @@ def ihex_line(address, record_type, data):
     line += "%02X\n" % ((256-check)%256)
     return line
 
-def ihex_parse(line):
-    assert line.startswith(":")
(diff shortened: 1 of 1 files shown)


debug/gdbserver.py
```diff
@@ -51,16 +51,25 @@ def ihex_line(address, record_type, data):
...
+        # data with 32-bit address
+        # Any higher bits were chopped off.
+        address = int(line[4:12], 16)
+        for i in range(6, count+1):
+            data += "%c" % int(line[2*i:2*i+2], 16)
+        # Ignore the checksum.
+        return 3, address, data
+    elif typ == 'S7':
+        # ignore execution start field
+        return 7, 0, 0
```
cgsfv: I would like to see an else clause.

debug/gdbserver.py
```diff
@@ -291,17 +300,22 @@ class MemTestBlock(GdbTest):
...
+            temporary_file.write(ihex_line(i * self.line_length, 0, line_data))
+        temporary_file.flush()
+        return data
+
+    def test(self):
+        a = tempfile.NamedTemporaryFile(suffix=".ihex")
+        data = self.write(a)
 
         self.gdb.command("shell cat %s" % a.name)
+        self.gdb.command("monitor riscv reset_delays 50")
```
cgsfv: Is reset_delays a new subcommand in openocd?

timsifive: Yes, added in https://github.com/riscv/riscv-openocd/pull/344 (which hasn't merged yet).

debug/gdbserver.py
timsifive: Done.

PR #177, opened 2019-01-26
Fix comments for shift amount.

takeoverjp:
Thank you for the very usefull test repository.

For `sll`, `srl`, `sra` insts, we have to consider the some low bits of rs2 as shift amount.
And according to the RISC-V spec 2.2, the bit width is

- 5 bits (in RV32)
- 6 bits (in RV64)

Current implementation has some unclear comments.
So I'd like to clarify them with this pull-req.


aswaterman:
And thank you for the helpful improvements to the comments!

takeoverjp:
Thank you for your prompt action!

Issue #158, opened 2018-09-06
breakpoint.S assumes the trigger is hardwired to breakpoint


tommythorn:
The test checks first that there's a breakpoint in the trigger (which I also find a bit strange, why is that a requirement?), the it overwrites it in line 32:
```
  li a0, MCONTROL_M | MCONTROL_EXECUTE
  csrw tdata1, a0
```
However note that the type isn't set (the top nibble is zero) and thus cleared.  It passes on spike because spike hardwires the trigger to 2.

aswaterman:
It’s not strange; it’s just a byproduct of no one else submitting better tests. Thanks for being the exceedingly rare counterexample.

tommythorn:
Does that mean that you agree that we should remove the test for existing breakpoint trigger (right before the code I changed)?  I can prep the change.

NB: I haven't even made my way through the rest of the test and probably will have other questions.
NB2: Still owe you more cases for the W variants.

aswaterman:
I don't think we should delete the test.  I think we should modify it to read back the CSR after writing it, and skip the test if trigger type 2 didn't stick.

aswaterman:
Re: the OT point: while we aren't doing this for riscv-tests yet, the continuous-integration support makes the github flow worthwhile.  It's nice when proposed patches have been regression-tested before they're even code-reviewed.

tommythorn:
Blush.  My change was incomplete.  First I should have stripped off the masking four lines down (now it always skip).  Furthermore the same change should have been applied in quite a few more places.
I'd recommend reverting it (I'm sorry).  Given my full change is so much longer, I think it merits a bit of discussion about what we are testing and what we are assuming.

aswaterman:
No problem, @tommythorn, I've reverted it for the time being.

PR #188, opened 2019-04-20
fix for #159 #158

neelgala:
This change will fix the issues pointed out in #159 and #158. 
The test no longer assumes the reset "type" of mcontrol to be 2 and thus explicitly defines it while updating tdata1.

Also tdata1 is read-back to see if the particular mcontrol fields are valid for the implementation else skip the test-part.


aswaterman:
Thanks, merged (but with the commented-out code deleted)

PR #195, opened 2019-07-03
Use work area in spike-1 to cover CRC algorithm.

timsifive:


timsifive:
Work areas are never required. They're only ever an optimization that OpenOCD can use if an appropriate area is configured.

Previously the work area was only used to speed up flashing, which spike doesn't simulate in any case. Now the work area is also used to speed up CRC computation, which happens when we ask gdb to `compare-sections` after a download. This change means that in that case we'll now use the optimized code path to get coverage of that. I didn't change it everywhere so we still get coverage of the non-optimized code also.

PR #200, opened 2019-07-30
Support RV32E. Fixed #198

### LINKED ISSUE #198: Support RV32E
https://github.com/riscv/riscv-tests/pull/196 It just supports the debug target. But the tests itself `test_macros.h` does not. https://github.com/riscv/riscv-tests/blob/master/isa/macros/scalar/test_macros.h

Does it support RISCV Test for RV32E in few month?

colin4124:


aswaterman:
I was concerned about the overlap with `a4` in the AMO/LR/SC tests, but those are using the `TEST_CASE` macros directly, so it all seems to work out.

Issue #206, opened 2019-09-20
debug: Path not search for executables (contrary to readme)


darius-bluespec:
According to the readme (https://github.com/riscv/riscv-tests/tree/readme/debug), certain executables "should be in the user's path".  However, those executable are not searched for in the user's path.  They are expected to be in $RISCV/bin/.

timsifive:
Good point. Which behavior do you think is preferable?

darius-bluespec:
I think the path is preferable.  Using $RISCV is an unnecessary nuisance, to me anyway.

PR #208, opened 2019-09-23
Look for binaries in $PATH.

timsifive:
Instead of relying on $RISCV.

Using $RISCV was common in the early days, but nowadays many tools are simply installed alongside the rest of the system.


timsifive:
Fixes #206

PR #220, opened 2019-11-23
Fix typo that makes rv32ud test rv32uf

TheThirdOne:
I ran into this when trying to test out doubles as I just implemented them, but it was very confusing to see that `fcvt.w.s` was showing up despite it not being present in the 64 bit double tests.

### DIFF
--- a/isa/rv32ud/fcvt_w.S
+++ b/isa/rv32ud/fcvt_w.S
@@ -4,4 +4,4 @@
 #undef RVTEST_RV64UF
 #define RVTEST_RV64UF RVTEST_RV32UF
 
-#include "../rv64uf/fcvt_w.S"
+#include "../rv64ud/fcvt_w.S"


isa/rv32ud/fcvt_w.S
```diff
@@ -4,4 +4,4 @@
 #undef RVTEST_RV64UF
 #define RVTEST_RV64UF RVTEST_RV32UF
```
jrtc27:
```suggestion
#define RVTEST_RV64UD RVTEST_RV32UD
```
?

isa/rv32ud/recoding.S
```diff
@@ -4,4 +4,4 @@
 #undef RVTEST_RV64UF
 #define RVTEST_RV64UF RVTEST_RV32UF
```
jrtc27:
```suggestion
#define RVTEST_RV64UD RVTEST_RV32UD
```
?

isa/rv32ud/fcvt_w.S
TheThirdOne: All of the files in rv32ud have the same #define, if that is incorrect, they should all be changed. I didn't know if that was a mistake or not so I didn't change it.

isa/rv32ud/recoding.S
TheThirdOne: See above

TheThirdOne:
This has been sitting here for a while. I would like to either get it merged in, a review that suggests I make additional changes, or an outright rejection.

isa/rv32ud/fcvt_w.S
abukharmeh: Hello, I think the #define RVTEST_RV64UF RVTEST_RV32UF part is not mistake, because the original define enables the floating point unit, but if you define it as RVTEST_RV64UD , the floating unit would not be enabled. The macro definition is in riscv_test.h, but the include is likely a mistake !

PR #223, opened 2019-12-05
Improve parallellism in debug test Makefile

timsifive:
Now each test is an individual make target, so you can get the most out
of however many cores you have. On my 12-core system, `make` went from
2m45s to 42s, and `make all` went from `3m25s` to `2m39s`.

If you have few cores, this change may actually slow things down a bit,
because ExamineTarget is run for every gdbserver.py invocation.

### DIFF
--- a/debug/Makefile
+++ b/debug/Makefile
@@ -12,9 +12,6 @@ all-tests: spike32 spike32-2 spike32-2-rtos spike32-2-hwthread \
 
 all:	pylint all-tests
 
-foo:
-	echo $(foreach test, $(TESTS), spike$@.$(test).run)
-
 run.%:
 	echo $@
 	$(GDBSERVER_PY) \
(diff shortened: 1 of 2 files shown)


debug/Makefile
```diff
@@ -11,19 +12,27 @@ all-tests: spike32 spike32-2 spike32-2-rtos spike32-2-hwthread \
 
 all:	pylint all-tests
 
-# Target to check all the multicore options.
-multi-tests: spike32-2 spike64-2-rtos spike32-2-hwthread
-
-pylint:
-	pylint --rcfile=pylint.rc `git ls-files '*.py'`
+foo:
```
SiFiveGregS: Just want to double check the intention was to keep "foo" as the name of that particular make target, or whether that was a placeholder name or temporary scaffolding (not that it really matters, but figured I'd ask).

timsifive: Oh, that was just a test while making this work. Thanks for noticing. I'll remove it.

PR #224, opened 2019-12-10
benchmarks: Simplify TLS initialisation

jrtc27:
The symbols used to query the size of .tdata and .tbss need not be thread-local themselves; instead, make them linker script-provided non-thread-local symbols.


aswaterman:
LGTM

Issue #115, opened 2018-02-20
Does the mi test for scall ever execute its test specific trap handler?


gameboo:
Hello. I believe that the `mi` test for `scall` is never reaching its test specific trap handler.

From what I understand in `riscv_test.h` (in the `p` target at least), the generic `trap_vector` handler catches `ecall`s and ends the test, presumably as a way to identify that we came from a call to a `RVTEST_PASS`/`RVTEST_FAIL` macro which sets up some register and performs an `ecall`. In the case where the exception cause is not coming from an `ecall`, the general trap handler falls through to the `mtvec_handler` test specific trap handler (which calls the test macros in the case of the `scall` test).

At least in the specific case of the `mi` test for `scall`, it seems to me that `mtvec_handler` cannot be reached since the "tested" `ecall` will first be caught as an `ecall` by the generic trap handler, preventing execution to to reach the test specific handler that would call the `RVTEST_PASS`/`RVTEST_FAIL` macro. I believe I can observe this behavior in an instruction trace from `spike`, but I might be missing something?

If that is of any use, I identified this while trying to run the tests with modified `RVTEST_PASS`/`RVTEST_FAIL` macros that simply access some CSRs mapped to terminate a simulation. All `ui` and `mi` tests were happy with that except for the `scall` one.

PR #245, opened 2020-02-20
Fixed bug in scall test when running in machine mode

Nils-TUD:
Since `stvec_handler` was defined to `mtvec_handler` _after_ using `RVTEST_CODE_BEGIN`, the macro still uses `stvec_handler`. This symbol does not exist in the end, so that it yields zero. For that reason, `mtvec` (`stvec` in the macro) is not set (see the `beqz` in line 156 of `env/p/riscv_test.h`). This has the consequence that the `ecall` in the test jumps to the normal `trap_vector`, which just jumps to `write_tohost` to report an error.

I can only guess what the intention of the test is, but I assume that the `ecall` executed in user mode should go to machine mode and jump to `stvec_handler`. Therefore, I'm setting `mtvec` manually in the test.

The patch works both for rv64mi and rv64si on gem5.

As far as I can see, this is the same issue as reported in #115.

### DIFF
--- a/isa/rv64si/scall.S
+++ b/isa/rv64si/scall.S
@@ -30,6 +30,10 @@ RVTEST_CODE_BEGIN
   li t1, CAUSE_USER_ECALL
 
 #ifdef __MACHINE_MODE
+  # register mtvec_handler
+  la t0, stvec_handler
+  csrw mtvec, t0
+
   # If running in M mode, use mstatus.MPP to check existence of U mode.
   # Otherwise, if in S mode, then U mode must exist and we don't need to check.
   li t0, MSTATUS_MPP


aswaterman:
What's confusing me is that the test works fine on Spike.  Is there any chance this patch is working around a bug in gem5 or something?

I've attached a log of the (successful) execution of these tests on Spike, using the HEAD of master of both Spike and this repo.  Can you let me know where they diverge from the gem5 execution?

[rv64mi-p-scall.log](https://github.com/riscv/riscv-tests/files/4232501/rv64mi-p-scall.log)
[rv64si-p-scall.log](https://github.com/riscv/riscv-tests/files/4232502/rv64si-p-scall.log)

Nils-TUD:
Oh, that's indeed my fault. Turns out that I misunderstood the way success/failure is reported and, astonishingly, this *only* caused a problem in this particular case. That is, without my change, gem5 behaves exactly like Spike, but I interpreted that as an error, not success :) Sorry for that.

That being said, I would still argue that my patch is useful. Because, no offense, but the scall test in machine mode is a bit weird. Primarily because `stvec_handler` is redefined to `mtvec_handler`, but the only purpose of that is to make the linker not find the symbol `stvec_handler`. `mtvec_handler` is in the binary, but never used. In other words, the test behaves completely different when executing in machine mode than when executing in supervisor mode. I don't see a good reason for that.

Therefore I think, explicitly setting `mtvec` to `mtvec_handler` makes it much more clear what's going on and makes the test behave (almost) the same in machine and supervisor mode.

aswaterman:
Glad we got to the bottom of it.  (I'm not going to argue that the intent of this test is particularly clear. PRs that add comments are welcome :))

PR #246, opened 2020-02-21
scall: make the intention of the test in machine mode more clear

Nils-TUD:


aswaterman:
So it seems this PR unintentionally relies on some assumptions about the behavior of the `RVTEST` macros that we didn't intend to hold.  So, for example, it fails with SiFive's internal versions of those macros.  Since the main purpose of this PR was making the intent clearer, I'm going to proceed with the following compromise: revert the PR, but add comments to the old version of the test to explain its admittedly confusing behavior.  Thanks again for helping.

PR #250, opened 2020-03-03
enable rv32e compatability by replacing reg x29 with reg x7

cedric-orban:
This allows the rv32ui tests to be run on rv32e cores.


aswaterman:
LGTM

PR #254, opened 2020-03-06
Don't assume reset state of mscratch

pdonahue-ventana:
The test currently assumes that mscratch is reset to 0 because it does a read before (concurrent with) the first write and it tests the read value.  The architecture has no such reset requirement.  This reorders test cases so the first access to mscratch is a write only.

### DIFF
--- a/isa/rv64si/csr.S
+++ b/isa/rv64si/csr.S
@@ -46,8 +46,8 @@ RVTEST_CODE_BEGIN
 #endif
 #endif
 
-  TEST_CASE(15, a0,         0, csrrwi a0, sscratch, 0; csrrwi a0, sscratch, 0xF);
   TEST_CASE(16, a0,         0, csrw sscratch, zero; csrr a0, sscratch);
+  TEST_CASE(15, a0,         0, csrrwi a0, sscratch, 0; csrrwi a0, sscratch, 0xF);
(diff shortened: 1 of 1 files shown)


aswaterman:
I don’t see the problem. The code “ csrrwi a0, sscratch, 0; csrrwi a0, sscratch, 0xF” does in fact read the uninitialized scratch register into a0, as you said, but then it immediately overwrites a0, so the uninitialized value in a0 is never tested.

pdonahue-ventana:
You're right and my description was wrong.

The thing that really motivated this is that the current code sequence doesn't play well with any sort of check that watches register file contents.  That could be any sort of instruction-by-instruction signature generation or comparison that compares one implementation (RTL or SW simulator) to another.  Or even single-stepping on silicon using an external debugger where you compare the x registers to a reference model after each instruction.  In some cases, you'll (for one instruction) get X in a register and in 2-state simulations you'll get random garbage.  Either would be bad for any comparison.

aswaterman:
Yeah, I figured that might be the reason.  This PR is fine, but the more general cosimulation-mismatch-due-to-initial-state problem probably needs to be solved outside of the tests themselves.  (Like how we're initializing the GPRs in the test environments.)

PR #255, opened 2020-03-10
Setup a multilevel page table to avoid misaligned superpages caused by variable DRAM_BASE

cedric-orban:
The single-level page table setup works fine if DRAM_BASE is greater than or equal to 0x4000_0000, however if DRAM_BASE is below this (e.g. 0x2000_0000), the PPN LSBs in the single PTE are set and the core traps because of a misaligned superpage.

To avoid this, I created a multilevel page table specific to sv39 (3 levels). In this setup, `satp` points to`superpage_table_1`, which points to `superpage_table_2`, which points to one of the the leaf PTEs.

### DIFF
--- a/isa/rv64si/dirty.S
+++ b/isa/rv64si/dirty.S
@@ -10,6 +10,10 @@
 #include "riscv_test.h"
 #include "test_macros.h"
 
+#if (DRAM_BASE >> 30 << 30) != DRAM_BASE
+# error This test requires DRAM_BASE be SV39 superpage-aligned
+#endif 
+
 RVTEST_RV64M
 RVTEST_CODE_BEGIN
 
--- a/isa/rv64si/dirty.S
+++ b/isa/rv64si/dirty.S
@@ -15,7 +15,7 @@ RVTEST_CODE_BEGIN
 
   # Turn on VM
   li a0, (SATP_MODE & ~(SATP_MODE<<1)) * SATP_MODE_SV39
-  la a1, superpage_table_1
+  la a1, page_table_1
   srl a1, a1, RISCV_PGSHIFT
   or a1, a1, a0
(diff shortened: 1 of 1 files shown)


aswaterman:
I get the point, but why don't we just change the code to `#define ALIGNED_DRAM_BASE ((DRAM_BASE + (1 << 22) - 1) >> 22 << 22)` and then use `ALIGNED_DRAM_BASE` everywhere?

It's OK if systems that can't spare the extra few MiB of memory can't run this particular test.  Not all tests need be applicable to all platforms, and most platforms that need virtual memory have ample DRAM these days.

(BTW, I used 2^22 because of RV32's 4 MiB megapagespages.)

cedric-orban:
I agree that I was misguidedly chasing smaller code size, but I'm not sure I follow your alternative solution, mainly because `DRAM_BASE == ALIGNED_DRAM_BASE` for some values of `DRAM_BASE` (e.g. multiples of 0x1000_0000).

aswaterman:
I'm actually going to retract my suggestion.  It is fine if this test is not-applicable for some platforms.  We can just statically fail the test if DRAM_BASE isn't superpage-aligned.

```
#if ((DRAM_BASE) >> 22 << 22) != DRAM_BASE
# error This test requires DRAM_BASE be superpage-aligned
#endif 
```

This is a very reasonable constraint in practice for systems with virtual memory, and for the rare other system, you can just not run this test.

cedric-orban:
Ok, I think it should be

```
#if (DRAM_BASE >> 30 << 30) != DRAM_BASE
# error This test requires DRAM_BASE be superpage-aligned
#endif 
```

in this case because the test targets sv39.

aswaterman:
Good point, my mistake.

PR #258, opened 2020-03-16
Add a test case rs = rd to jalr.S

takahirox:
Hi all, thank you for the very nice project!

I made [RISC-V processor emulator in Rust](https://github.com/takahirox/riscv-rust) and succeeded in [running xv6-riscv on the emulator on browser by compiling the emulator to WebAssembly](https://takahirox.github.io/riscv-rust/index.html). ([xv6](https://pdos.csail.mit.edu/6.828/2019/xv6.html) is UNIX V6 rewritten by MIT for x86 and [xv6-riscv](https://github.com/mit-pdos/xv6-riscv) is its RISC-V port.) riscv-tests helped me so much to verify my CPU emulator implementation.

While developing my emulator, I noticed that JALR test doesn't cover rs = rd case. At the beginning of my emulator development xv6 couldn't run due to rs = rd case bug while I passed JALR riscv-tests.

So this PR adds rs = rd case to jalr.S. I confirmed with this change

- no error in rv32/64ui-p/v-jalr test on spike
- no error in my latest emulator implementation
- error in my old emulator implementation (see below) as expected

This is my first contribution to riscv-tests. Please let me know anything wrong in my PR.

**The problem in detail**

JALR instruction sets x[rd] to pc and sets pc to x[rs] + imm. In my old emulator implementation I wrote

```
// pseudo code
// wrong implementation
1: x[rd] <= pc;
2: pc <= x[rs] + imm;
```

but it works wrongly where rs = rd because x[rs (=rd)] is overridden with pc in line 1 before updating pc from x[rs] in line 2. I should had rewritten to

```
// pseudo code
// correct implementation
1: tmp <= pc
2: pc <= x[rs] + imm;
3: x[rd] <= tmp
```

But the current JALR riscv-tests doesn't take care rs = rd case so it doesn't detect the bug I mentioned.

This PR adds rs=rd case to detect such a bug.

```
test_3:
  li  TESTNUM, 3
  la  t0, target_3

  jalr t0, t0, 0
linkaddr_3:
  j fail

target_3:
  la  t1, linkaddr_3
  bne t0, t1, fail
```

In the new test case, for example the wrong implementation above will go to linkaddr_3 because

```
1: x[rd] <= pc;
2: pc <= x[rs] + imm; # pc <= pc + 0
```

If I remember correctly, I found similar problems in swap type instructions tests (I have bad memory so I forgot which ones... I will check later). So I'll make follow up PRs for them if this PR looks ok and gets merged.


aswaterman:
Interesting.  I guess this is the kind of bug that would likely never show up in a real HW implementation, since pipeline design tends to preclude the possibility of intra-instruction WAR hazards.  But I can easily see how it would show up in a software simulator.  Thanks for the contribution.

PR #261, opened 2020-03-17
Improve address translation tests

timsifive:
Now only test modes that are actually supported by hardware, and turn off PMP.

Now it passes on HiFive Unleashed if you hack the config file not to expose hart 0. (Otherwise it'll correctly report the test is not applicable.) There's still the deeper OpenOCD issue regarding memory accesses always happening through the first hart in a `-rtos hwthread` setup. I'll start looking at that, since it'll probably be a problem some day.

### DIFF
--- a/debug/testlib.py
+++ b/debug/testlib.py
@@ -1070,6 +1070,17 @@ class GdbTest(BaseTest):
 
         self.gdb.select_hart(self.hart)
 
+    def disable_pmp(self):
+        # Disable physical memory protection by allowing U mode access to all
+        # memory.
+        try:
+            self.gdb.p("$pmpcfg0=0xf")  # TOR, R, W, X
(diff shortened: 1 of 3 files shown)


timsifive:
@SiFiveGregS, can you take a look at this?

debug/testlib.py
```diff
@@ -1070,6 +1070,17 @@ def parkOtherHarts(self):
 
         self.gdb.select_hart(self.hart)
 
+    def disable_pmp(self):
+        # Disable physical memory protection by allowing U mode access to all
+        # memory.
+        try:
+            self.gdb.p("$pmpcfg0=0xf")  # TOR, R, W, X
```
SiFiveGregS: Ah, so top-of-range in the pmp0cfg slot implicitly treats bottom of range as address 0x0.  Hadn't noticed that before, but makes sense under the principle of least surprise.

PR #271, opened 2020-03-30
Change slen to a value that spike supports.

timsifive:


timsifive:
@SiFiveGregS, can you review this?

PR #295, opened 2020-10-02
Modify PMP benchmark to automatically detect granularity

Moschn:
This PR modifies the PMP benchmark to automatically detect the granularity implemented by the core. The granularity is detected according to the privilege spec:

> Software may determine the PMP granularity by writing zero to `pmp0cfg`, then writing all ones to `pmpaddr0`, then reading back `pmpaddr0`. If G is the index of the least-significant bit set, the PMP granularity is 2G+2 bytes.

This was tested on Spike and Ariane.

### DIFF
--- a/benchmarks/pmp/pmp.c
+++ b/benchmarks/pmp/pmp.c
@@ -8,6 +8,7 @@
 #include "util.h"
 
 volatile int trap_expected;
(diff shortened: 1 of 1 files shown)


aswaterman:
LGTM, thanks.

PR #297, opened 2020-10-07
Expose registers on all harts in openocd cfgs

sobuch:
This should solve the issue raised at https://github.com/riscv/riscv-openocd/issues/543


timsifive (approved):
Thank you!

PR #299, opened 2020-10-19
fence_i.S : use registers present on rv32e

SandeepRajendran:
The fence_i test was using register 16 which isn't present on rv32e designs. Use register 6 instead to support rv32e.


SandeepRajendran:
@aswaterman please review.

aswaterman:
LGTM.

PR #301, opened 2020-11-06
add zfh (float16) test case and related macros

chihminchao:


chihminchao:
@aswaterman  Is it acceptable ?

aswaterman:
LGTM, thank you, @chihminchao!

PR #309, opened 2020-12-15
Revive and expand invalid read test.

timsifive:
This test would have found
https://github.com/riscv/riscv-openocd/issues/559.

### DIFF
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -238,18 +238,25 @@ class MemTest64(SimpleMemoryTest):
     def test(self):
         self.access_test(8, 'long long')
 
-# FIXME: I'm not passing back invalid addresses correctly in read/write memory.
-#class MemTestReadInvalid(SimpleMemoryTest):
-#    def test(self):
-#        # This test relies on 'gdb_report_data_abort enable' being executed in
-#        # the openocd.cfg file.
-#        try:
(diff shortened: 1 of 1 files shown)


debug/gdbserver.py
```diff
@@ -238,18 +238,25 @@ class MemTest64(SimpleMemoryTest):
...
+        # This test relies on 'gdb_report_data_abort enable' being executed in
+        # the openocd.cfg file.
+        try:
+            self.gdb.p("*((int*)0x%x)" % bad_address)
+            assert False, "Read should have failed."
+        except testlib.CannotAccess as e:
+            assertEqual(e.address, bad_address)
+        self.gdb.stepi()    # Don't let gdb cache register read
+        assertEqual(self.gdb.p("*((int*)0x%x)" % good_address), 0xabcdef)
+        assertEqual(self.gdb.p("$s0"), 0x12345678)
```
SiFiveGregS, line 258: Is the point of checking s0 that s0 is known to normally get saved/used/restored internally by the debugger's memory access mechanisms, so this is a check that the original s0 value in fact gets correctly preserved?

SiFiveGregS (approved):

debug/gdbserver.py
timsifive: Yes. Specifically, the bug involved s0 (aka fp) being corrupted when a memory access failed.

PR #318, opened 2021-01-09
Don't rely on the implementation-specific WFI time limit

pdonahue-ventana:
The architecture says:
> When TW=1, then if WFI is executed in any less-privileged mode, and it does not complete within an implementation-specific, bounded time limit, the WFI instruction causes an illegal instruction exception. The time limit may always be 0, in which case WFI always causes an illegal instruction exception in less-privileged modes when TW=1.

This test was testing the behavior in the second sentence.  But if the implementation-specific time limit is not 0 then the first sentence applies and the WFI instruction is not guaranteed to trap.  There's no architectural way to detect whether the time limit is 0 other than doing what this test does and seeing what happens.  That's pointless so I removed that whole test case (and the prior bad3 undefined instruction trap which existed only to perform setup for this bad4 case).


aswaterman:
Yeah, I think the test was correct when written, but the spec was later revised.

PR #320, opened 2021-01-21
Smoketest that vl and vtype can be modified.

timsifive:


SiFiveGregS (approved):
I suppose if failed writes of 0 to vector registers starts being seen with some future hw implementations, then this test can be adjusted then (or ignored)

Issue #321, opened 2021-01-30
Erroneously recursive code generated for non-recursive memset() in syscalls.c with -O2


quxfred:
https://github.com/riscv/riscv-tests/blob/e75931826097c741f274212a49302e39ca597168/benchmarks/Makefile#L40

![image](https://user-images.githubusercontent.com/60102550/106360636-83b1e600-6311-11eb-9e15-4ef08d3a36c2.png)

Note the recursion marked in orange in above disassembly is supposed to perform the else-branch at common/syscalls.c:406. But this just causes a stack overflow. Adding  -fno-tree-loop-distribute-patterns or -ffreestanding got rid of this error. It looks having to do with a recurring GCC bug that attempts some obscure transformation on memset-like code.

See also:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56888
https://gcc.gnu.org/legacy-ml/gcc-help/2019-10/msg00060.html
https://gcc.gnu.org/legacy-ml/gcc-help/2019-10/msg00076.html

Regards,
Freddie

aswaterman:
See
https://github.com/riscv/riscv-gcc/issues/197

quxfred:
So this has been known for months. I'll close this issue then.

Should the work-around be in the Makefile?

aswaterman:
It should be in the Makefile, but someone needs to take the time to figure out how to do it without being too disruptive to performance. -fno-builtin is too heavy of a hammer because it destroys dhrystone performance for other reasons. Jim’s suggestion on the ticket I linked to is probably the right answer, but that needs to be validated, by making sure that dhrystone’s use of string copy, struct copy, etc. are still inlined.

(Note the reason it hasn’t been fixed yet is that older versions of the compiler don’t exhibit this problem. Not that that fact helps with your situation, I realize.)

aswaterman:
I’ll reopen the ticket since the problem isn’t actually fixed.

quxfred:
Just made a quick-and-dirty patch in PR #322. Sorry I don't know how to put it under this issue.

aswaterman:
I double-checked that the solution proposed in https://github.com/riscv/riscv-gcc/issues/197 works on older compilers, too, and it's a cleaner fix.  I'll commit that and close this issue.

PR #322, opened 2021-01-31
A quick-fix work-around for issue #321

quxfred:
Turn affected memset() C code into inline assembly to avoid GCC meddle with it.

### DIFF
--- a/benchmarks/common/syscalls.c
+++ b/benchmarks/common/syscalls.c
@@ -394,6 +394,35 @@ void* memcpy(void* dest, const void* src, size_t len)
 
 void* memset(void* dest, int byte, size_t len)
 {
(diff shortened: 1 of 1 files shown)


quxfred:
A little explanation of this patch.

As raised in the issue, GCC miscompiles the standalone memset() in syscalls.c. Passing a global compiler flag can stop GCC misbehaving on memset(), but would destroy dhrystone performance. To avoid the latter, I took the correct assembly code as GCC spit out with the right set of flags, and turned the whole memset() body into inline assembly.

The change is protected with a macro USE_INLINE_ASSEMBLY, which is also confined locally within the function, so nothing else is expected to change. Once this bug is fixed on GCC, the C body can be restored. I didn't bother hand-crafting the assembly, but feel free to do that if anyone has the time!

PR #323, opened 2021-02-01
Make mtvec_handler's address to be multiple of 16

muhammadumar-lm:
Setting mtvec_handler's address to be multiple of 16. This avoids overriding of mtvec_handler's address by writing it into mtvec CSR for certain configs.


aswaterman:
`.align 2` should suffice here.

PR #333, opened 2021-04-08
Add FreeRTOS smoke tests.

timsifive:
Make sure that OpenOCD can access all threads in a FreeRTOS binary on
single-hart RV32 and RV64.


debug/gdbserver.py
```diff
@@ -1700,6 +1700,60 @@ def test(self):
...
+
+        values = {}
+        for thread in threads:
+            assertNotIn("No Name", thread[1])
+            self.gdb.thread(thread)
+            assertEqual(self.gdb.p("$zero"), 0)
+            output = self.gdb.command("info reg sp")
+            assertIn("ucHeap", output)
+            self.gdb.command("info reg mstatus")
+            values[thread.id] = self.gdb.p("$s11")
```
SiFiveGregS, line 1748: Is there anything special about the choice of s11 as the register here?  (e.g. is the demo running under FreeRTOS known to use s11 for some specific purpose?).  Or was the choice of specific register here relatively unimportant for the functioning of the test case?

SiFiveGregS (approved):

debug/gdbserver.py
timsifive: It's just one that I deemed to be unlikely to be clobbered by the stepi. In the future I want to write a test that changes more registers, and confirms that those changes actually affect the code being debugged.

PR #334, opened 2021-05-05
Test daisy chained homogeneous spike instances.

timsifive:
For OpenOCD, this means we're checking that we can talk to multiple TAPs.


SiFiveGregS (approved):
The remote bitbang chaining proxy looks handy, it's amazing how concise that ends up being, in Python

PR #335, opened 2021-05-09
Add the method to build riscv-tests on ARM in README

Yuwenfeng2019:
Recently,  when I tried to build riscv-tests in order to test the Rocket Chip that built on my Raspberry Pi 4(abbreviated as "RPi4" from now on), it was found that the original guide of riscv-tests seems only focus on X86, and does not apply for ARM. After several attempts, finally, I successfully built riscv-tests on RPi4 and used it to do some tests against Rocket Chip, everything seems OK. 
Thus this pull request is made for as a reference for those who are going to develop RISC-V on ARM hardware like RPi4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,12 @@ install path, and that the riscv-gnu-toolchain package is installed.
     $ make
     $ make install
 
+If you are on an ARM machine, e.g., Raspberry Pi 4 with Fedora Spin 33
+XFCE AArch64, you may use the similar configure arguments as below to
+cross-compile riscv-tests:
+./configure --prefix=$RISCV/target --build=aarch64-redhat-linux --host=riscv64-unknown-elf --target=riscv64-unknown-elf
+The above target triplets depends on your host system and the toolchain prefix according to $RISCV.
+
 The rest of this document describes the format of test programs for the RISC-V
 architecture.
 


aswaterman:
Nothing about the instructions is actually x86-specific.  It sounds like maybe we need to update some configure scripts to cope with platform differences.  I'll leave this open for now, in case it helps other people trying to solve this problem, but it's not the right fix.

Yuwenfeng2019:
> Nothing about the instructions is actually x86-specific. It sounds like maybe we need to update some configure scripts to cope with platform differences. I'll leave this open for now, in case it helps other people trying to solve this problem, but it's not the right fix.

Yes!  Thanks for reviewing it!
Actually, according to my practice of building riscv-tests on RPi4, there is no need to change the source code of the project , but just use different configure arguments as this pull request pointed out, that's why I only made a change to the README as a "note" to the developers who are used to work non-x86 machines for their daily work(nowadays, more and more ARM chips are coming to desktop or server system, such like Apple M1, AWS Graviton2, and so on). 
Let me further clarify how I need to deal with riscv-tests on ARM. Recently, I built and tested Firrtl,Treadle, and Chisel on RPi4, and it does work, then I also built the Rocket Tools directly on RPi4 with some modifications, and set RISCV environment variable  according to it. Then I followed the official guide and some online document(e.g., https://www.programmersought.com/article/7539888504/) to try to build a HelloWorld program with riscv-tests for testing the emulators of Rocket Chip, and found that the original configure script for riscv-tests that described in README does not work on ARM, so finally, the solution to this issue result in this pull request--I successfully built the riscv-tests and HelloWorld with specified target triplets for the configuration according to the system of RPi4 and the toolchain prefix in RISCV, and passed the related tests within the above documents.

PR #336, opened 2021-05-10
Fixes for illegal.S to support Bare-SMode and sbreak.S & scall.S to support CLIC mode.

slametr-sifive:
illegal.S:
- After the test enters supervisor mode, check if paging is supported.
- If paging is NOT supported (i.e. Bare S-mode), jump to a new section of code that checks the following:
-- SFENCE.VMA causing illegal instruction trap regardless of TVM.
-- Access to SATP does not trap.
-- Jump to the same TSR check as regular S-mode
-- End test

sbreak.S & scall.S:
- Before checking for scause, check if the core is in CLIC-mode (mtvec[1]).
- If we're in CLIC-mode, mask off scause bits[(XLEN-1):8] before checing its value.
- Otherwise, don't mask off any scause bits as in the original test.

### DIFF
--- a/isa/rv64mi/illegal.S
+++ b/isa/rv64mi/illegal.S
@@ -60,6 +60,18 @@ msip:
   # Make sure WFI doesn't trap when TW=0.
   wfi
 
+  # Check if paging is supported (Set SUM & MXR and read it back)
+  and t0, t0, zero
+  li t0, (SSTATUS_SUM | SSTATUS_MXR)
+  csrc sstatus, t0
+  and t1, t1, zero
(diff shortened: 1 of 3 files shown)


slametr-sifive:
@aswaterman , for some reason I cannot add anybody as reviewers.  Could you help with reviewing and merging this PR?  Thanks.

aswaterman:
@slametr-sifive I think you need to be a member of the corresponding github organization (or something like that) in order to request reviews.  But @-mentioning people works fine, too.

The SUM/MXR trick looks sound for now, but I'm not certain it will work once the enhanced PMP proposal is adopted.  That's OK; we can cross that bridge later.

PR #337, opened 2021-05-12
Fix for rv64mi/sbreak and rv64mi/scall

slametr-sifive:
Fix for rv64mi/sbreak and rv64mi/scall that I broke in my previous commit:
Added "#define stvec mtvec" under __MACHINE_MODE ifdef.


slametr-sifive:
@aswaterman , sorry I broke rv64mi/sbreak and rv64mi/scall with my previous commit.
It's just a one line fix. Thanks.

PR #348, opened 2021-07-18
Debug tests: catch write to nonexistent trigger registers in entry.S

Wren6991:
The reset handler in `entry.S` attempts to clear all hardware triggers before starting the test. An implementation can be compliant without implementing the trigger module, in which case this access will trap and then spin forever in the trap handler. This breaks all software-based debug tests on targets which don't implement the trigger module. I hit this when bringing up the debug tests on my soft core.

This patch adds some trap-catch code around the trigger CSR accesses in the reset handler. Since the purpose of this routine is to clear any triggers which may be set, the code serves no purpose on targets with no trigger module, so the failure can be safely ignored, and startup can continue.

### DIFF
--- a/debug/programs/entry.S
+++ b/debug/programs/entry.S
@@ -71,6 +71,9 @@ handle_reset:
   addi  t0, t0, -1
   bnez  t0, 1b
 
+  # Catch trap in case trigger module is not implemented
+  la t2, 2f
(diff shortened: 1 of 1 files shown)


timsifive (approved):
Looks good! Thank you for fixing this.

PR #349, opened 2021-07-18
Add a test for Svnapot

daniellustig:


aswaterman:
Just skimmed, but LGTM.

PR #353, opened 2021-07-22
Fix #352

### LINKED ISSUE #352: napot.S sets U bit in non-leaf PTEs
The privileged spec says:
> For non-leaf PTEs, the D, A, and U bits are reserved for future standard use and must be cleared by software for forward compatibility.

Step 2 of the translation algorithm says:
> Let pte be the value of the PTE at address a+va.vpn[i]×PTESIZE. (For Sv32, PTESIZE=4.) If accessing pte violates a PMA or PMP check, or if any reserved bits in pte are set, raise an access-fault exception corresponding to the original access type.

I don't see anything in Svnapot that changes that.

These lines in napot.S set the U bit in non-leaf PTEs:
https://github.com/riscv/riscv-tests/blob/master/isa/rv64si/napot.S#L61
https://github.com/riscv/riscv-tests/blob/master/isa/rv64si/napot.S#L81

Therefore, the instructions that attempt to use the NAPOT translation should be getting page faults.  This seems like a test bug and probably also a spike bug.

@daniellustig

daniellustig:
Thanks to @pdonahue-ventana for pointing this out

### DIFF
--- a/isa/rv64ssvnapot/napot.S
+++ b/isa/rv64ssvnapot/napot.S
@@ -58,7 +58,7 @@ RVTEST_CODE_BEGIN
   la a0, page_table
   srl a0, a0, 12
   sll a0, a0, PTE_PPN_SHIFT
-  li a1, PTE_V | PTE_U
+  li a1, PTE_V
   or a0, a0, a1
 
   # Level 1 PTE address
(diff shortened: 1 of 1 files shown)


pdonahue-ventana (commented):
Looks good.

PR #358, opened 2021-08-24
masking maskmax in breakpoints test

pentin-as:
Currently in rocket-chip, reading tdata1 will return the value "4" in the maskmax field.
This happens because of this line:
https://github.com/chipsalliance/rocket-chip/blob/7f214d564a57d3b1fa63e67a42a64629e76d53a2/src/main/scala/rocket/Breakpoint.scala#L30
Thus, the breakpoint.S test was always skipped and did not work.


pentin-as:
Screenshot from 2021-08-25 09-03-10

PR #364, opened 2021-11-10
Create DisconnectTest.

timsifive:
Disconnects from gdb, and then reconnects, making sure that didn't
change any of the registers.

This test will start passing when
https://github.com/riscv/riscv-openocd/pull/661 merges.


SiFiveGregS (approved):
Interesting; somehow I wasn't aware of the existence of GDB's "disconnect" command (was already familiar with "detach" but ok so that one resumes the target, which clearly isn't what we want for this test).

PR #367, opened 2022-01-04
Add gdb.interact() for debug tests.

timsifive:
This feature lets you easily interact with the gdb after the test has
run to a certain point.


SiFiveGregS (approved):
looks like a handy feature

PR #369, opened 2022-02-08
Debug test to check that stepping doesn't inappropriately switch to Thread 1

SiFiveGregS:
A new test to check for regression/reappearance of the issue fixed in riscv-openocd commit f6ffede8b62e7092141d54e75238fde1ee00d0c2

### DIFF
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -1801,8 +1801,12 @@ class StepThread2Test(GdbTest):
         output = self.gdb.command("thread 2")
         if "Unknown thread" in output:
             raise TestNotApplicable
-        output = self.gdb.stepi()
-        assertNotIn("Switching to Thread 1", output)
+        before = self.gdb.command("thread")
+        self.gdb.stepi()
+        after = self.gdb.command("thread")
+        # make sure that single-step doesn't alter
+        # GDB's conception of the current thread
+        assertEqual(before, after)
 
 
 parsed = None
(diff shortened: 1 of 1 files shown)


timsifive (commented):
What does this test do against the targets/RISC-V/spike32-2.py? In that case you have two instances of gdb connected to a 2-hart target. Each individual gdb would only see one thread.

More generally, it would be better to check for registers or something than checking for the gdb message, but that's a lot more complicated to test. As it is this test is a good improvement and I'm happy to merge it once we figure out the deal with spike32-2.py.

SiFiveGregS:
For Spike32-2 target, the test fires a "not applicable" result, because the bug that we were seeing happens only in cases where a single GDB instance sees the target as having more than 1 thread.   For targets that were showing the bug (e.g. spike32-2-hwthread and spike64-2-hwthread) detecting the issue with register values is tricky because the register values after the stepi are correct for the thread that was actually stepped.  The only wrong aspect was a missing thread ID in the stop reply, which led GDB to arbitrarily consider thread 1 to be current when that was simply not true.  Maybe the test case could be adjusted to issue a "thread" command (with no arguments) before and after the stepi, and then assert that the output of both of those "thread" commands is identical (or at least the numeric thread ID is unchanged).

timsifive:
I like the idea of testing the output of `thread`. That feels more robust than relying on a warning message that might change or not exist in some versions of gdb.

SiFiveGregS:
Thanks, I'll try that.

SiFiveGregS:
Verified that test passes (or not applicable for non-hwthread topologies) with riscv-openocd commit f6ffede8b, and that test fails as expected for pre-bug-fix commit 7d91f639b (or not applicable for non-hwthread topologies).

timsifive (approved):
Looks great. Thank you!

PR #374, opened 2022-03-04
Add coremark benchmark

jiegec:
Add coremark benchmark from https://github.com/eembc/coremark.

Example output:

```shell
$ spike benchmarks/coremark.riscv
2K performance run parameters for coremark.
CoreMark Size    : 666
Total ticks      : 35417288
Total time (secs): 0.354173
Iterations/Sec   : 282.347988
ERROR! Must execute for at least 10 secs for a valid result!
Iterations       : 100
Compiler version : GCC12.0.1 20220129 (experimental)
Compiler flags   : -O2 -ffast-math -fno-common -fno-builtin-printf -fno-tree-loop-distribute-patterns
Memory location  : STACK
seedcrc          : 0xe9f5
[0]crclist       : 0xe714
[0]crcmatrix     : 0x1fd7
[0]crcstate      : 0x8e3a
[0]crcfinal      : 0x988c
Errors detected
mcycle = 35417303
minstret = 35417308
```


aswaterman:
My preference is actually to nuke the benchmarks, rather than adding to them. They are a holdover from teaching courses at UC Berkeley. RISC-V International should not be implicitly endorsing benchmarks.

PR #376, opened 2022-03-12
Loosen failing store-conditional test

colluca:
The ISA specification mandates that:
> If the SC.W fails, the instruction does not write to memory, and it writes a
nonzero value to rd.

Test 2 in `rv64ua-p-lrsc` however checks that the return value is exactly 1.
https://github.com/riscv-software-src/riscv-tests/blob/3e2bf06b071a77ae62c09bf07c5229d1f9397d94/isa/rv64ua/lrsc.S#L28-L33

This PR proposes to loosen the constraint by checking not equivalency to 1, but rather inequality to 0.
This allows cores implementing the looser option, while complying to the specification, to make use of the `riscv-tests` suite.

A notable example being the CVA6 core [failing](https://github.com/openhwgroup/cva6/pull/838) on this test after recent changes.

### DIFF
--- a/isa/macros/scalar/test_macros.h
+++ b/isa/macros/scalar/test_macros.h
@@ -8,7 +8,8 @@
 # Helper macros
 #-----------------------------------------------------------------------
 
-#define MASK_XLEN(x) ((x) & ((1 << (__riscv_xlen - 1) << 1) - 1))
+#define XLEN_MASK ((1 << (__riscv_xlen - 1) << 1) - 1)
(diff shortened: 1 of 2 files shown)


aswaterman:
“The failure code with value 1 is reserved to encode an unspecified failure.  Other failure codes are reserved at this time”

Standard RISC-V cores must report either 0 or 1. It is a feature that this text explicitly checks for the value 1, rather than accepting any nonzero value.

PR #380, opened 2022-04-12
Add EbreakTest.

timsifive:
Confirm correct behavior when somebody bakes an ebreak instruction into
their code.

### DIFF
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -1735,6 +1735,37 @@ class VectorTest(GdbSingleHartTest):
         assertIn("_exit", output)
         assertEqual(self.gdb.p("status"), 0)
 
+class EbreakTest(GdbSingleHartTest):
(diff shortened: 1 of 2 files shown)


debug/gdbserver.py
```diff
@@ -1735,6 +1735,37 @@ def test(self):
         assertIn("_exit", output)
         assertEqual(self.gdb.p("status"), 0)
 
+class EbreakTest(GdbSingleHartTest):
+    """Test that we work correctly when somebody puts an ebreak directly into
+    their code."""
+    compile_args = ("programs/ebreak.c", )
```
SiFiveGregS, line 1741: Where does the file "ebreak.c" live?

timsifive: Forgot to commit. Good catch!

SiFiveGregS (approved):

PR #384, opened 2022-05-16
Test semihosting_fileio

timsifive:
In the original test, confirm that stdout data ends up in the OpenOCD
log.
In the new test, with `arm semihosting_fileio` enabled, confirm that
stdout data ends up in gdb's CLI.

This test requires https://github.com/riscv/riscv-openocd/pull/699.

### DIFF
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -952,16 +952,37 @@ class Semihosting(GdbSingleHartTest):
         assertIn("Breakpoint", output)
         assertIn("_exit", output)
         assertEqual(self.gdb.p("status"), expected_result)
+        return output
 
     def test(self):
-        """Sending gdb ^C while the program is running should cause it to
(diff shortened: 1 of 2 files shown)


debug/gdbserver.py
```diff
@@ -966,6 +965,28 @@ def test(self):
...
+
+    def test(self):
+        temp = tempfile.NamedTemporaryFile(suffix=".data")
+
+        self.gdb.b("main:begin")
+        self.gdb.c()
+        self.gdb.p('filename="%s"' % temp.name, ops=3)
+        output = self.exit()
+        # stdout should end up in gdb's CLI
+        assertIn("Do re mi fa so la ti do!", output)
```
SiFiveGregS, line 985: solfège output in a test program, nice

SiFiveGregS (approved):

PR #387, opened 2022-05-25
Address trigger has higher priority than alignment

timsifive:
Used this test to confirm that
https://github.com/riscv-software-src/riscv-isa-sim/pull/1013 works
right.

### DIFF
--- a/isa/rv64mi/breakpoint.S
+++ b/isa/rv64mi/breakpoint.S
@@ -98,6 +98,51 @@ RVTEST_CODE_BEGIN
   bnez a2, fail
 
 2:
+  # Test that address breakpoints have higher priority than unaligned loads.
+  li TESTNUM, 12
+  li a0, (2 << (__riscv_xlen - 4)) | MCONTROL_M | MCONTROL_LOAD | MCONTROL_STORE
+  csrw tdata1, a0
+  # Skip if breakpoint type is unsupported.
+  csrr a1, tdata1
+  bne a0, a1, 2f
+  la a2, data1 + 1
+  la a3, data3
+  csrw tdata2, a2
+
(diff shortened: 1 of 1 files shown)


isa/rv64mi/breakpoint.S
```diff
@@ -97,6 +97,51 @@ RVTEST_CODE_BEGIN
...
+
+  # Test that address breakpoints have higher priority than unaligned stores.
+  li TESTNUM, 16
+
+  la a2, data1 + 1
+  csrw tdata2, a2
+
+  # Slow path because CSR was *just* written. (This is spike-specific.)
+  sw x0, (a2)
+  beqz a2, fail
```
aswaterman, line 133: Presumably this branch should be removed since the store can't possibly write `x2`?

isa/rv64mi/breakpoint.S
```diff
@@ -97,6 +97,51 @@ RVTEST_CODE_BEGIN
...
+
+  # Normal aligned store, so that the next time we'll hit the fast path.
+  li TESTNUM, 17
+  sw x0, (a3)
+
+  # Fast path
+  li TESTNUM, 18
+  sw x0, (a2)
+  beqz a2, fail
```
aswaterman, line 142: Same here.  Or I guess if you're trying to make sure the store didn't write memory, you'd need to store a nonzero value and `lb` a byte of it back to see if it's still zero.

isa/rv64mi/breakpoint.S
timsifive: Good point. Definitely a cut and paste issue. I just removed the checks, since if the trigger didn't match then you end up with an unaligned exception which will cause the test to fail.

PR #391, opened 2022-06-02
Add test cases for unaligned lw

timsifive:


aswaterman:
Does this test pass on spike both with and without misaligned accesses enabled?

timsifive:
It does.

aswaterman (approved):

PR #395, opened 2022-06-07
Test misaligned stores

timsifive:
This passes `make -C isa run` against spike built with and without `--enable-misaligned`.

If this organization is good, I'll do something similar for stores.


isa/macros/scalar/test_macros.h
```diff
@@ -12,9 +12,9 @@
 
 #define TEST_CASE( testnum, testreg, correctval, code... ) \
 test_ ## testnum: \
+    li  TESTNUM, testnum; \
```
aswaterman, line 15: I'm glad this change only ended up affecting a handful of tests; I would've expected it to have greater blast radius.

PR #397, opened 2022-06-09
Test misaligned stores.

timsifive:
Assume that misaligned loads work correctly.

Passes `make -C isa run` against spike (both with and without misaligned
support).


aswaterman:
LGTM.  Thanks!

aswaterman (approved):
LGTM. Thanks!

PR #402, opened 2022-07-03
Debug: MemorySampleMixed: Disable 64-bit memory sampling on 32-bit harts

Wren6991:
Currently the MemorySampleMixed test unconditionally attempts to sample an 8-byte memory region. This bails out at this check in OpenOCD, on my implementation which supports `sbaccess` = 0, 1, 2:

https://github.com/riscv/riscv-openocd/blob/8488e4e863f6a465189d7dd3e6d94ccc4c493eb4/src/target/riscv/riscv-013.c#L2165-L2169

It seems reasonable to require 64-bit SBA for sampling a 64-bit variable, so that your samples are atomic. However, there are no requirements (AFAIK) on the sizes supported by SBA, and if all of your harts have XLEN=32 then it's likely this is your maximum SBA size too.

This patch removes monitoring of the 64-bit `i64` variable when hart XLEN is less than 64.


debug/gdbserver.py
```diff
@@ -887,7 +887,8 @@ def test(self):
 class MemorySampleMixed(MemorySampleTest):
     def test(self):
         addr = {}
-        for i, name in enumerate(("j", "i32", "i64")):
+        test_vars = ("j", "i32") + ("i64",) * (self.hart.xlen >= 64)
```
timsifive, line 890: Just to make the code more clear:
```suggestion
        test_vars = ["j", "i32"]
        if self.hart.xlen >= 64:
            test_vars.append("i64")
```

timsifive (changes requested):
This makes sense to me.

debug/gdbserver.py
Wren6991: You're right, this is much clearer. I force-pushed this change.

timsifive (approved):

PR #406, opened 2022-07-22
Ignore `mip` and `time` in DisconnectTest.

timsifive:
These can change at any time and that is OK.


SiFiveGregS (approved):
Looks good; it makes sense that mip and time can have volatile values across debugger disconnect/reconnect.

PR #225, opened 2019-12-11
benchmarks: Various fixes for compiling with Clang

jrtc27:

PR #409, opened 2022-09-15
fix compile for Clang

geekLucian:
When trying to compile riscv-tests with clang,  several incompatible compilation issues between clang and gcc are met, such as:
- https://github.com/riscv-software-src/riscv-tests/pull/225
- clang doesn't support immediates like `0f:7fc00000`
- clang doesn't support bf16 until https://github.com/llvm/llvm-project/commit/e4888a37d36780872d685c68ef8b26b2e14d6d39, but it only works for X86 yet, and will not affect current release of clang (i.e. clang-14)
- GNU as let `.weak` override `.globl`, while clang lets the last directive win https://reviews.llvm.org/D90108


aswaterman:
Did you verify this all still works with GCC?

PR #410, opened 2022-09-26
rv64ui test misaligned load/store data

ingallsj:
### What was changed?
Starting from `rv64mi/ma_addr`:
1. Remove the machine-level code allowing for a trap, so that this test can run in User Mode.
2. Add scenarios for misaligned crossing 32- and 64-byte boundaries.

### What should be reviewed
@timsifive building on the tests added for misaligned loads https://github.com/riscv-software-src/riscv-tests/pull/395 and stores https://github.com/riscv-software-src/riscv-tests/pull/397.

### How this was tested or validated
passes when run internally

### DIFF
--- a/isa/rv64ui/Makefrag
+++ b/isa/rv64ui/Makefrag
@@ -12,6 +12,7 @@ rv64ui_sc_tests = \
 	jal jalr \
 	lb lbu lh lhu lw lwu ld \
 	lui \
+	ma_data \
 	or ori \
 	sb sh sw sd \
 	sll slli slliw sllw \
--- a/isa/rv64ui/ma_data.S
+++ b/isa/rv64ui/ma_data.S
@@ -1,0 +1,387 @@
+# See LICENSE for license details.
+
+#*****************************************************************************
+# ma_data.S
+#-----------------------------------------------------------------------------
+#
+# Test misaligned ld/st data.
+# Based on rv64mi-ma_addr.S
+#
+
+#include "riscv_test.h"
+#include "test_macros.h"
+
+RVTEST_RV64U
+RVTEST_CODE_BEGIN
+
+  la s0, data
+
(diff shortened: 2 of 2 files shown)


aswaterman:
@ingallsj Is there any hope of sharing the main body of these tests?  If not, that's OK, too.

ingallsj:
What do you mean by "main body"?  There's a lot of exact and subtly shifting offsets and constants.

I suppose we can always refactor... these just came from a different commit history.

aswaterman:
Sorry, ambiguous pronoun.  I meant the body of `ma_addr` and `ma_data`.  At first glance, it looks like the body of this test is an expanded version of that one.

ingallsj:
Yes, I think we can, I'll work on doing that.

ingallsj:
Reading more closely, the `ma_addr` checks on misaligned stores are loads that the first and last bytes are zero or not.  The `ma_data` checks misaligned stores using misaligned loads.

So, they're different enough that I'd like to not spend the effort combining them right now.

aswaterman:
@ingallsj Makes sense to me.

aswaterman (approved):

rapiz1:
Hello guys I would like to know more context of this test.

According to the ISA manual
> Volume I: RISC-V Unprivileged ISA V20191213, p25
> Loads and stores where the effective address is not naturally aligned to the referenced datatype (i.e., on a four-byte boundary for 32-bit accesses, and a two-byte boundary for 16-bit accesses) have behavior dependent on the EEI.

The hardware may not support misaligned load&store and instead raise exceptions, as tested in `lw-unaligned` and `sw-unaligned`, `ma_addr`. But in this test I didn't see any exception handler.

Am I correct about the following conclusion? This test only applies to those who implement misaligned load&store, and failing to pass this test does not mean any violation of the spec or misfunctioning of the processor.

ingallsj:
Yes, your understanding is correct 👍

PR #411, opened 2022-09-26
zicbo test zero

ingallsj:
### What was changed?
1. Makefile for Zicbo* extensions tests.
2. Add a very simple test for CBO.ZERO: do one of them, checking all 64 bytes with 8 `ld` instructions.  Assume a 64-byte cache line size (Zic64b).  Execute the test in Machine mode, so we don't need to program the `envcfg.CBZE` bits.

After this works and is in, we can move on to Zicbom.

### What should be reviewed
@timsifive

### How this was tested or validated
not tested

### DIFF
--- a/isa/Makefile
+++ b/isa/Makefile
@@ -17,6 +17,7 @@ include $(src_dir)/rv64uzfh/Makefrag
 include $(src_dir)/rv64si/Makefrag
 include $(src_dir)/rv64ssvnapot/Makefrag
 include $(src_dir)/rv64mi/Makefrag
+include $(src_dir)/rv64mzicbo/Makefrag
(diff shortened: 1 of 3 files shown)


aswaterman:
@ingallsj Any reason we can't make this a U-mode test instead?  I prefer to make these privileged only if there's a good reason.

ingallsj:
I think we would need to start in M mode to set the `menvcfg.CBZE` bit, and the same in `henvcfg`/`senvcfg`, to execute in U mode.

aswaterman:
True.  I guess this is OK for now.

aswaterman (approved):

PR #413, opened 2022-10-04
Update testlib.py; remove ANSI escape sequences

YenHaoChen:
The control sequences (^[[?2004h and ^[[?2004l) occur after the gdb.command, which results in Exception fault. This commit removes the control sequences and strips out the blank lines (^M).


timsifive (approved):
Looks good. Thanks.

PR #414, opened 2022-10-04
update gdbserver.py; tolerance value of MemorySampleTest()

YenHaoChen:
The value seems to differ significantly from the previous one after the 'timestamp' string. The difference could be over the tolerance value. Thus, this commit reset the previous_value after the 'timestamp.'


timsifive (commented):
The point of this change to ensure that memory samples are happening "quickly." Unfortunately it's hard to define what quickly means. I've set tolerance based on running against spike on my computer. What target are you using where the tolerance is insufficient?

timsifive (approved):
Sounds good. Thanks.

PR #418, opened 2022-10-07
debug: On failure, disassemble close instructions.

timsifive:
`disassemble` shows the whole function which is usually too much. Instead just show the nearest instructions for some context.


timsifive:
Merged in #432.

PR #420, opened 2022-10-12
Get coverage of progbuf FPR accesses.

timsifive:
Using the new spike support merged in
https://github.com/riscv-software-src/riscv-isa-sim/pull/1109


SiFiveGregS (approved):
So I guess fp register access was previously being tested through the abstract command pathway only, and this option now gives an option for testing same accesses via PROGBUF; good idea.

timsifive:
That's right.

PR #422, opened 2022-10-20
Change memory address used in debug tests.

timsifive:
https://github.com/riscv-software-src/riscv-isa-sim/pull/889 put a UART at the address we were using in our 32-bit debug tests.


aswaterman (approved):
Oops!

PR #424, opened 2022-10-24
fix DebugFunctionCall after https://github.com/riscv-software-src/riscv-tests/commit/90d1637f8c934b218c0e0f98c942b4d60cef97e2

YenHaoChen:
The DebugFunctionCall fails after https://github.com/riscv-software-src/riscv-tests/commit/90d1637f8c934b218c0e0f98c942b4d60cef97e2.

Changing the default value seems to fix the issue.


timsifive:
This should pass with spike versions that have https://github.com/riscv-software-src/riscv-isa-sim/pull/1109 (recently added). Does your spike have that option?
This changes makes it so we lose the coverage that we gained in #420.

YenHaoChen:
Thank you. The case passed after pulling the newest spike.

PR #425, opened 2022-10-25
Specify trigger type=2 in trigger.S

YenHaoChen:
Previous tests implicitly assume triggers only support type=2. However, a trigger may support multiple types, i.e., type=15. This commit explicitly specifies type=2 in trigger.S to support type 15.

### DIFF
--- a/debug/programs/trigger.S
+++ b/debug/programs/trigger.S
@@ -40,6 +40,7 @@ write_loop:
         j       main_exit
 
 write_store_trigger:
+        /* 2<<60 is for RV64. 2<<28 is for RV32. That's safe because on RV64 bits 28 and 29 are 0. */
         li      a0, (2<<60) | (2<<28) | (1<<6) | (1<<1)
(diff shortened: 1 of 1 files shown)


debug/programs/trigger.S
```diff
@@ -40,14 +40,14 @@ write_loop:
         j       main_exit
 
 write_store_trigger:
-        li      a0, (1<<6) | (1<<1)
+        li      a0, (2<<60) | (2<<28) | (1<<6) | (1<<1)
```
timsifive, line 43:
```suggestion
        /* 2<<60 is for RV64. 2<<28 is for RV32. That's safe because on RV64 bits 28 and 29 are 0. */
        li      a0, (2<<60) | (2<<28) | (1<<6) | (1<<1)
```

timsifive (changes requested):

timsifive (approved):

PR #427, opened 2022-10-31
Fix PrivChange test address comparison.

timsifive:
Before it might fail incorrectly, because main was close to trap_entry.


SiFiveGregS (approved):
looks good; stronger and more specific test case than previously

PR #428, opened 2022-11-03
SvNNTest needs 32KB of RAM.

timsifive:
Also change the test itself to require less RAM than it did previously. (It had required more than 32KB.)


timsifive:
@SiFiveGregS can you take a look at this?

SiFiveGregS (approved):

PR #429, opened 2022-11-11
Add tests that make a hart unavailable while running

timsifive:
Use the `cease` instruction for this. It's unfortunately non-standard, but it does exist in Rocket so it's not crazy to use it here.

### DIFF
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -18,6 +18,7 @@ from testlib import assertGreater, assertRegex, assertLess
 from testlib import GdbTest, GdbSingleHartTest, TestFailed
 from testlib import TestNotApplicable, CompileError
 from testlib import UnknownThread
+from testlib import CouldNotReadRegisters
 
 MSTATUS_UIE = 0x00000001
 MSTATUS_SIE = 0x00000002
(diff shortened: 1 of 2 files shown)


debug/programs/entry.S
```diff
@@ -204,6 +204,18 @@ trap_entry:
...
+  .align 2
+precease:
+    // Loop a while so that OpenOCD might have confirmed the resume before the
+    // hart becomes unavailable.
+    li t1, 100000
+1:
+    addi t1, t1, -1
+    bnez t1, 1b
+cease:
+  .word 0x30500073  // cease
```
SiFiveGregS, line 216: So I guess Spike supports the CEASE instruction, that's good

SiFiveGregS (approved):

debug/programs/entry.S
aswaterman: It doesn't; it'll raise an illegal-instruction exception.

timsifive: Right. I've been running this against an FPGA image that does support the cease instruction.

SiFiveGregS: ah, yeah, got it, thanks!

timsifive:
Looks like I already used this branch for a PR that's no longer necessary. (Mostly merged in #434?)

PR #432, opened 2022-11-29
debug: Disassemble memory when a failure happens.

timsifive:
This gives you less noise in the log, and more chance of figuring out what code was actually executed.

### DIFF
--- a/debug/testlib.py
+++ b/debug/testlib.py
@@ -1217,7 +1217,7 @@ class GdbTest(BaseTest):
             return
         self.gdb.interrupt()
         self.gdb.command("info breakpoints", reset_delays=None)
-        self.gdb.command("disassemble", ops=20, reset_delays=None)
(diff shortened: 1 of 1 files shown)


debug/testlib.py
```diff
@@ -1217,7 +1217,7 @@ def postMortem(self):
             return
         self.gdb.interrupt()
         self.gdb.command("info breakpoints", reset_delays=None)
-        self.gdb.command("disassemble", ops=20, reset_delays=None)
+        self.gdb.command("x/20i $pc", ops=20, reset_delays=None)
```
SiFiveGregS, line 1220: Ah, so disassemble doesn't always give useful output if the pc is outside a known source-level function.  This new way looks better, for the situation.

SiFiveGregS (approved):

debug/testlib.py
timsifive: And it gives too verbose output if you're in a large function, when it disassembles the whole thing.

PR #433, opened 2022-11-30
Share exit() among more tests.

timsifive:


SiFiveGregS (approved):
The reduced duplication looks good.

PR #435, opened 2022-12-01
Update Makefile

ssayin:
Signed-off-by: Serdar Sayın <25414028+ssayin@users.noreply.github.com>

Hello, I attempted to compile your code using a compiler built with the help of the following script, where DEFAULTARCH is set to rv32imac: https://github.com/embecosm/embecosm-toolchain-releases/blob/master/stages/build-riscv32-gcc.sh#L14 . The build fails with errors stating that the zicsr extension is required. I suggest this tiny change to also ensure backwards compatibility.


ssayin:
Considering that isa/ also depends on this Makefile, this is probably not a good idea. The pull request is now closed.

PR #436, opened 2022-12-01
debug: Create CeaseMultiTest.

timsifive:
Confirm basic debug still works when other harts have been parked using a `cease` instruction. Check that the unavailable harts are inaccessible from gdb.

Add Gdb.expect()
Parse "unknown thread" error from gdb.

### DIFF
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -12,10 +12,12 @@ import re
 
 import targets
 import testlib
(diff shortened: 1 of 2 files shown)


SiFiveGregS (approved):
Good idea to save this off as a repeatable test case, I'm guessing this is similar to how the "hart becomes unavailable" stuff was tested during development.

timsifive:
Yes. The majority of my development is making the tests pass. :-)

PR #442, opened 2022-12-15
Add CeaseStepiTest and CeaseRunTest

timsifive:


debug/gdbserver.py
```diff
@@ -1832,6 +1833,51 @@ def test(self):
...
+        return self.hart.support_cease
+
+    def test(self):
+        self.gdb.b("main")
+        output = self.gdb.c()
+        assertIn("Breakpoint", output)
+        assertIn("main", output)
+
+        self.gdb.p("$pc=cease")
+        self.gdb.stepi(wait=False)
```
SiFiveGregS, line 1849: So the wait=False is introduced in order to be prepared for, and verify the presence of, asynchronous output that reports the available->unavailable change, whereas previously a stepi would be expected to complete silently in all cases.  Looks good to me.

SiFiveGregS (approved):

PR #445, opened 2022-12-30
Pass --misaligned flag to Spike to run ISA tests

aswaterman:
...because the ma_data test requires this feature.

Don't merge until https://github.com/riscv-software-src/riscv-isa-sim/pull/1206 is merged.


jerryz123:
This should get merged now.

PR #449, opened 2023-01-20
Fix ma_fetch test for writable misa.C

jerryz123:
Fixes bug introduced by #444


aswaterman:
Can you explain the rationale? Perhaps it’s obvious, but I’m on my phone and haven’t re-read the whole test.

jerryz123:
The code should be saving, writing, and then restoring `mtvec` to catch the case where clearing `misa.C` does not set `IALIGN=32`. I wrote this using `mepc` erroneously, but didn't catch it after spike dropped support for writable `misa.C`.

aswaterman (approved):
Makes sense.

PR #451, opened 2023-02-02
Fix EtriggerTest on multi-hart targets.

timsifive:
Need to set the etrigger on the hart we're running the test against.


timsifive:
@SiFiveGregS Can you review this?

SiFiveGregS (approved):

PR #462, opened 2023-03-02
Stop using linker-script for tests

dimitry-:
My understating is that it can be replaced with -Wl,-Ttext to produce the same result. It also avoids leaving headers out of PT_LOAD segment as well as having RWE flag for one of the segments.

### DIFF
--- a/isa/Makefile
+++ b/isa/Makefile
@@ -58,11 +58,11 @@ vpath %.S $(src_dir)
 define compile_template
 
 $$($(1)_p_tests): $(1)-p-%: $(1)/%.S
-	$$(RISCV_GCC) $(2) $$(RISCV_GCC_OPTS) -I$(src_dir)/../env/p -I$(src_dir)/macros/scalar -T$(src_dir)/../env/p/link.ld $$< -o $$@
+	$$(RISCV_GCC) $(2) $$(RISCV_GCC_OPTS) -I$(src_dir)/../env/p -I$(src_dir)/macros/scalar -Wl,-Ttext=0x80000000 $$< -o $$@
 $(1)_tests += $$($(1)_p_tests)
 
 $$($(1)_v_tests): $(1)-v-%: $(1)/%.S
-	$$(RISCV_GCC) $(2) $$(RISCV_GCC_OPTS) -DENTROPY=0x$$(shell echo \$$@ | md5sum | cut -c 1-7) -std=gnu99 -O2 -I$(src_dir)/../env/v -I$(src_dir)/macros/scalar -T$(src_dir)/../env/v/link.ld $(src_dir)/../env/v/entry.S $(src_dir)/../env/v/*.c $$< -o $$@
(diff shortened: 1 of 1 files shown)


aswaterman (changes requested):
This breaks RV64 tests on my system.  And I'd actually prefer to keep the linker script, because the additional dependence on the toolchain's built-in linker script is another thing that can break across toolchain upgrades.

```
$ make rv64ui-p-add.out
riscv64-unknown-elf-gcc -march=rv64g -mabi=lp64 -static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles -I./../env/p -I./macros/scalar -Wl,-Ttext=0x80000000 rv64ui/add.S -o rv64ui-p-add
spike --isa=rv64gc_zfh_zicboz_svnapot --misaligned rv64ui-p-add 2> rv64ui-p-add.out
make: *** [Makefile:53: rv64ui-p-add.out] Error 255

$ cat rv64ui-p-add.out
Access exception occurred while loading payload rv64ui-p-add:
Memory address 0x7ffff000 is invalid

$ riscv64-unknown-elf-readelf -Wl rv64ui-p-add

Elf file type is EXEC (Executable file)
Entry point 0x80000000
There are 2 program headers, starting at offset 64

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x000000 0x000000007ffff000 0x000000007ffff000 0x0016bc 0x0016bc R E 0x1000
  LOAD           0x0016c0 0x00000000800016c0 0x00000000800016c0 0x000048 0x000048 RW  0x1000

 Section to Segment mapping:
  Segment Sections...
   00     .text 
   01     .tohost
```

dimitry-:
> Memory address 0x7ffff000 is invalid

I think I need some education here - why is this address invalid?

dimitry-:
To elaborate a little bit on reasons for the change. Some platforms will likely require better security practices and prohibit rwe mappings. And unfortunately using linker script together with ld.bfd produces such segments. Another reason is that program headers are being left out of load segment and this is unusual (I have another PR to fix this in linker scripts here: https://github.com/riscv/riscv-test-env/pull/41).

aswaterman:
> I think I need some education here - why is this address invalid?

There's nothing fundamentally wrong with that address.  But the default Spike memory map has memory at 0x80000000 and above, which is why the linker script does what it does.

aswaterman:
> To elaborate a little bit on reasons for the change. Some platforms will likely require better security practices and prohibit rwe mappings. And unfortunately using linker script together with ld.bfd produces such segments. Another reason is that program headers are being left out of load segment and this is unusual (I have another PR to fix this in linker scripts here: [riscv/riscv-test-env#41](https://github.com/riscv/riscv-test-env/pull/41)).

I agree RWX mappings are a poor security practice, but these aren't normal programs; they're ISA test cases.  They're supposed to run on the bare metal.  So, while I agree with your general comment, I don't agree that it applies here.

dimitry-:
Running isa cases on bare metal is not the only use-case, we are planning to run them on the emulator. As for LOAD address issue - it is addressed here: https://github.com/dimitry-/riscv-tests/commit/8c0203a3f4f6e1bee67d6205e6dc129992f3c276, Unfortunately it no longer visible in this pull request since it is closed.

aswaterman:
The tests are already written in such a way that you can build and run them in different environments.  You can already use the existing source code and use a different Makefile/linker script/whatever.  I don't think that there is a problem that needs to be solved.

PR #463, opened 2023-03-16
Fix breakpoint.S failing when tcontrol is implemented

Wren6991:
`tcontrol.mte` is required to reset to 0 when implemented. When 0, triggers don't fire. This patch attempts to set `tcontrol.mte` before beginning the breakpoint test, skipping over the `tcontrol` access if it traps.

There weren't any encoding.h constants for tcontrol, so I raised https://github.com/riscv/riscv-test-env/pull/42 to add those. If that gets merged, I'll update this PR to use the correct symbolic constants.


aswaterman (approved):
LGTM.

PR #465, opened 2023-04-06
Make sure that sc after failure sc fails

mohanson:
Regardless of success or failure, executing an SC.W instruction invalidates any reservation held by this hart. So sc after failure sc must fails.

### DIFF
--- a/isa/rv64ua/lrsc.S
+++ b/isa/rv64ua/lrsc.S
@@ -97,7 +97,7 @@ TEST_CASE( 7, a1, 1, \
   addi a3, a0, 8; \
 1:lr.w a1, (a0); \
   sc.w a1, x0, (a3); \
-  bnez a1, 1b; \
+  beqz a1, 1b; \
   sc.w a1, x0, (a0)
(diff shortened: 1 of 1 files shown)


aswaterman (changes requested):
This test doesn't work.  It gets stuck in an infinite loop when executed on Spike because the first SC is to a different address than the LR.

mohanson:
I mistakenly used an old `lrsc` binary for testing. I have fixed this.

aswaterman:
I don't think this version is correct, either, even though it now passes on Spike.  The problem is that implementations are allowed to reserve more than one word of memory on an LR, so the SC might succeed on some implementations even though the addresses differ.

I made a version of this that does work in general, by piggy-backing on the previous test, which does contain an SC that we know will fail by construction.  https://github.com/riscv-software-src/riscv-tests/commit/394937eceb62bc4ba8ad5a4e916f0508d9803229

mohanson:
Your version is better, this PR no longer needs to open.

PR #466, opened 2023-04-06
Include Zicntr in Spike ISA string

aswaterman:
Spike no longer enables Zicntr by default, so turn it on explicitly.

cc @jerryz123


jerryz123 (approved):

aswaterman:
@jerryz123 can you approve again now that you have permissions on this repo?

aswaterman:
@jerryz123 oops, I think you need to accept the invite first, then try again.

jerryz123 (approved):

jerryz123:
There we go

PR #468, opened 2023-04-17
Add more bmarks

jerryz123:
Vector implementations are taken straight from the examples in the RVV 1.0 spec repo/


benchmarks/memcpy/memcpy.S
```diff
@@ -0,0 +1,111 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
```
aswaterman, line 1: I don't want to import GPL code into this repository.  Can you improve the `memcpy` in `syscalls.c`?  Unrolling the main loop, doing all the loads into local variables before doing the stores, should be enough to match the performance of this version, provided the input arrays are XLEN-bit aligned.

benchmarks/mt-memcpy/mt-memcpy.c
```diff
@@ -0,0 +1,68 @@
...
+//--------------------------------------------------------------------------
+// Main
+//
+// all threads start executing thread_entry(). Use their "coreid" to
+// differentiate between threads (each thread is running on a separate core).
+
+void thread_entry(int cid, int nc)
+{
+   // static allocates data in the binary, which is visible to both threads
+   static int results_data[DATA_SIZE];
```
aswaterman, line 46: Might want to change the data type to `long` or `uintptr_t` so that you don't end up with 4-byte alignment on RV64.

benchmarks/Makefile
```diff
@@ -32,14 +32,15 @@ bmarks = \
...
+	vec-memcpy \
 
 #--------------------------------------------------------------------
 # Build rules
 #--------------------------------------------------------------------
 
 RISCV_PREFIX ?= riscv64-unknown-elf-
 RISCV_GCC ?= $(RISCV_PREFIX)gcc
-RISCV_GCC_OPTS ?= -DPREALLOCATE=1 -mcmodel=medany -static -std=gnu99 -O2 -ffast-math -fno-common -fno-builtin-printf -fno-tree-loop-distribute-patterns -march=rv$(XLEN)gc -mabi=lp64d
+RISCV_GCC_OPTS ?= -DPREALLOCATE=1 -mcmodel=medany -static -std=gnu99 -O2 -ffast-math -fno-common -fno-builtin-printf -fno-tree-loop-distribute-patterns -march=rv$(XLEN)gcv -mabi=lp64d
```
aswaterman, line 46: I was going to suggest that we want to turn off autovectorization so that the C code continues to run on implementations without the V extension, but it appears that only happens at `-O3`, so this should be OK.

benchmarks/vec-saxpy/vec-saxpy.S
```diff
@@ -0,0 +1,30 @@
+    .text
+    .balign 4
+    .global vec_saxpy
+# void
+# vec_saxpy(size_t n, const double a, const double *x, const double *y, double* z)
```
aswaterman, line 5: This is actually DAXPY (i.e. double precision), not SAXPY.  I'd be fine with changing either correcting the name or changing the data type.

aswaterman (changes requested):
OK with me after my concerns are addressed.

benchmarks/memcpy/memcpy.S
jerryz123: Thanks, its now the unrolled syscalls.c version.

benchmarks/vec-saxpy/vec-saxpy.S
jerryz123: Oops. Fixed

aswaterman (approved):

PR #469, opened 2023-05-01
Update OpenOCD cfg files to new syntax

timsifive:
We were using a variety of deprecated commands.

The driving force behind this was the new way to use `expr{}` as the old way no longer works with mainline OpenOCD.

### DIFF
--- a/debug/targets/RISC-V/spike-1.cfg
+++ b/debug/targets/RISC-V/spike-1.cfg
@@ -1,8 +1,8 @@
-adapter_khz     10000
+adapter speed     10000
 
-interface remote_bitbang
-remote_bitbang_host $::env(REMOTE_BITBANG_HOST)
-remote_bitbang_port $::env(REMOTE_BITBANG_PORT)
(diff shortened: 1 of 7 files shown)


JanMatCodasip (approved):
Checked visually - looks all right.

SiFiveGregS (approved):
I wonder if the "expr" behavioral change is specific to the JimTcl flavor of Tcl; looks like the "full" Tcl 8.6 implementation of expr will still concatenate multiple arguments.  But enclosing argument(s) in curly braces is still recommended even there for most cases.  (Maybe JimTcl now goes one step further to encourage that style).

timsifive:
> I wonder if the "expr" behavioral change is specific to the JimTcl flavor of Tcl; looks like the "full" Tcl 8.6 implementation of expr will still concatenate multiple arguments. But enclosing argument(s) in curly braces is still recommended even there for most cases. (Maybe JimTcl now goes one step further to encourage that style).

I didn't really dig into this, but I got an error without the {} so made the change.

PR #477, opened 2023-06-13
Add a way to exclude tests by specifying an exclusion file

MarekVCodasip:
This patch adds a way to specify a yaml file which specifies either for each target individually or for all targets to exclude tests. Example file format is included in excluded.yaml.example.

### DIFF
--- a/debug/excluded.yaml.example
+++ b/debug/excluded.yaml.example
@@ -1,0 +1,17 @@
+# This is an example file showing the exclusion list format.
+# It can be passed to gdb_server.py using the --exclude-tests argument.
+
+
+# Tests excluded for all targets
+all:
+  - SimpleF18Test
+
+# Tests excluded for the "spike32" target only
+spike32:
+  - MemTest32
+  - PrivRw
+  - Sv32Test
+
+# Tests excluded for the "spike64" target only
+spike64:
+  - UserInterrupt
--- a/debug/testlib.py
+++ b/debug/testlib.py
@@ -11,6 +11,7 @@ import time
 import traceback
 
 import pexpect
+import yaml
 
 print_log_names = False
 real_stdout = sys.stdout
@@ -947,6 +948,32 @@ class PrivateState:
     def __exit__(self, _type, _value, _traceback):
         self.gdb.pop_state()
 
+
+def load_excluded_tests(excluded_tests_file, target_name):
+    result = []
+    if excluded_tests_file is None or len(excluded_tests_file) == 0:
+        return result
+
+    target_excludes = {}
+    with open(excluded_tests_file) as file:
+        raw_data = yaml.safe_load(file)
+        for (target, test_list) in raw_data.items():
+            if not isinstance(test_list, list):
+                raise ValueError(f"Target {target!r} does not contain a test list", excluded_tests_file, test_list)
+            if not all(isinstance(s, str) for s in test_list):
+                raise ValueError(f"Not every element in the target test list {target!r} is a string",
+                                 excluded_tests_file, test_list)
+
+        target_excludes.update(raw_data)
+
+    if target_name in target_excludes:
+        result += target_excludes[target_name]
+    if "all" in target_excludes:
+        result += target_excludes["all"]
+
+    return result
+
+
 def run_all_tests(module, target, parsed):
     todo = []
     for name in dir(module):
@@ -984,6 +1011,9 @@ def run_all_tests(module, target, parsed):
         elif not examine_added:
             todo.insert(0, ("ExamineTarget", ExamineTarget, None))
             examine_added = True
+
+    excluded_tests = load_excluded_tests(parsed.exclude_tests, target.name)
+    target.skip_tests += excluded_tests
 
     results, count = run_tests(parsed, target, todo)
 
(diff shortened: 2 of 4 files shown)


MarekVCodasip:
Hey @timsifive, can you please take a look at this.

debug/excluded.yaml.example
```diff
@@ -0,0 +1,17 @@
...
+
+# Tests excluded for the "spike32" target only
+spike32:
+  - MemTest32
+  - PrivRw
+  - Sv32Test
+
+# Tests excluded for the "spike64" target only
+spike64:
+  - UserInterrupt
```
JanMatCodasip, line 17: @MarekVCodasip Detail - it looks like there's a missing newline at the end of this text file.

(Explanation: https://stackoverflow.com/a/729795)

JanMatCodasip (commented):
@MarekVCodasip One more improvement suggestion: You can also add the following to the Makefile to simplify the use of the exclusion lists:

```diff
diff --git a/debug/Makefile b/debug/Makefile
index 06f7d9d..b412e58 100644
--- a/debug/Makefile
+++ b/debug/Makefile
@@ -23,7 +23,8 @@ run.%:
                --isolate \
                --print-failures \
                --sim_cmd $(RISCV)/bin/$(RISCV_SIM) \
-               --server_cmd $(RISCV)/bin/openocd
+               --server_cmd $(RISCV)/bin/openocd \
+               $(if $(EXCLUDE_TESTS),--exclude-tests $(EXCLUDE_TESTS))
 
 # Target to check all the multicore options.
 multi-tests: spike32-2 spike32-2-hwthread

```

The use would then be:

```bash
EXCLUDE_TESTS=my_exclusion_list.yml make all
```

debug/excluded.yaml.example
MarekVCodasip: Added

debug/testlib.py
```diff
@@ -947,6 +948,31 @@ def __enter__(self):
...
+        return result
+
+    target_excludes = {}
+    with open(excluded_tests_file) as file:
+        raw_data = yaml.safe_load(file)
+        for test_list in raw_data.values():
+            if not isinstance(test_list, list):
+                raise ValueError(excluded_tests_file, test_list)
+            if not all(isinstance(s, str) for s in test_list):
+                raise ValueError(excluded_tests_file)
```
timsifive, line 964: Can we provide the user a meaningful error message here?

timsifive (changes requested):
Looks like a good feature, that could be used in the future to maintain known lists of failing tests as well.

JanMatCodasip (approved):

debug/testlib.py
```diff
@@ -947,6 +948,33 @@ def __enter__(self):
...
+        return result
+
+    target_excludes = {}
+    with open(excluded_tests_file) as file:
+        raw_data = yaml.safe_load(file)
+        invalid_format_message = "Invalid format for the exclusion file. Please take a look at the " \
+                                 "\"excluded.yaml.example\" file as an example of the test exclusion file format."
+        for test_list in raw_data.values():
+            if not isinstance(test_list, list):
+                raise ValueError(invalid_format_message, excluded_tests_file, test_list)
```
timsifive, line 964: Specifically here, can you tell the user something like "list of tests wasn't a list".

Even better would be to do `for (target, test_list) in raw_data.items()` above so the error can include the name of the target the list belongs to.

debug/testlib.py
```diff
@@ -947,6 +948,33 @@ def __enter__(self):
...
+            if not all(isinstance(s, str) for s in test_list):
+                raise ValueError(invalid_format_message, excluded_tests_file)
```
timsifive, line 966: And here, "not every element in the test list is a string"

timsifive (changes requested):

debug/testlib.py
MarekVCodasip: Done

debug/testlib.py
MarekVCodasip: Added with the target name

debug/testlib.py
```diff
@@ -947,6 +948,32 @@ def __enter__(self):
...
+    result = []
+    if excluded_tests_file is None or len(excluded_tests_file) == 0:
+        return result
+
+    target_excludes = {}
+    with open(excluded_tests_file) as file:
+        raw_data = yaml.safe_load(file)
+        for (target, test_list) in raw_data.items():
+            if not isinstance(test_list, list):
+                raise ValueError(f"Target \"{target}\" does not contain a test list", excluded_tests_file, test_list)
```
timsifive, line 962:
```suggestion
                raise ValueError(f"Target {target!r} does not contain a test list", excluded_tests_file, test_list)
```

debug/testlib.py
```diff
@@ -947,6 +948,32 @@ def __enter__(self):
...
+            if not all(isinstance(s, str) for s in test_list):
+                raise ValueError(f"Not every element in the target test list \"{target}\" is a string",
```
timsifive, line 964:
```suggestion
                raise ValueError(f"Not every element in the target test list {target!r} is a string",
```

debug/testlib.py
timsifive: This way python will add the quotes if it is a string, and also escape newlines or whatever else might end up in the string if the data is malformed.

timsifive (changes requested):
Last nitpicky comments.

timsifive (approved):

PR #478, opened 2023-06-15
debug: optimize the FreeRtosTest case.

Du-Chao:
To make sure the rtos module of OpenOCD works well.


timsifive (commented):
This looks like a good improvement.

We can merge this after https://github.com/riscv/riscv-openocd/pull/865 merges.

timsifive (approved):

PR #479, opened 2023-06-15
Add --target-timeout to debug test script.

timsifive:
I'm using this to greatly reduce the timeout when I'm reproducing a failure I know is going to time out.


timsifive:
@SiFiveGregS Can you take a look at this? Should be trivial.

PR #486, opened 2023-06-30
change warning check in RepeatReadTest

en-sc:


en-sc:
Related to https://github.com/riscv/riscv-openocd/pull/871

debug/gdbserver.py
```diff
@@ -912,10 +916,15 @@ def test(self):
         output = self.gdb.command(
             f"monitor riscv repeat_read {count} 0x{mtime_addr:x} 4")
         values = []
-        for line in output.splitlines():
-            # Ignore warnings
-            if line.startswith("Batch memory"):
-                continue
```
timsifive, line 918: I still see an error at https://github.com/riscv/riscv-openocd/pull/871/files#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63R3574 that would be caught by this. Are you sure it's correct to remove this warning?

en-sc: It's a `LOG_TARGET_DEBUG`, so it's not reported to user (in a gdb session).

en-sc:
Old warnings are also supported now, in accordance with [this comment](https://github.com/riscv/riscv-openocd/pull/871#issuecomment-1623962270) by @timsifive.

timsifive (approved):

PR #489, opened 2023-07-07
debug: Test OpenOCD behavior when harts become unavailable, using new spike mechanism

timsifive:
These tests won't pass until https://github.com/riscv/riscv-openocd/pull/879 merges.


SiFiveGregS (approved):
Some really helpful and intricate test infrastructure here.

timsifive:
These tests sometimes fail (depending on which hart is randomly selected) in the multi-spike scenario. Instead of fixing that, I'm changing it so they don't run there. The fix is somewhat complex, the test takes forever to run (15+ minutes), and the additional coverage is negligible.

timsifive:
Fixed a but that happened when running commands through the OpenOCD CLI when `-d` is passed to OpenOCD.

SiFiveGregS (approved):

PR #492, opened 2023-07-14
Add unratified Zacas extension

ved-rivos:
This PR adds tests for the unratified fast-track Zacas extension that introduces the following instructions:
- amocas.w for RV32 and RV64
- amocas.d for  RV32 and RV64
- amocas.q for RV64 only
The specification is here: https://github.com/riscv/riscv-zacas
The amocas instructions are emitted using a .word directive pending assembler support.


aswaterman:
@ved-rivos did you make sure these can run on Spike?

ved-rivos:
I have tested them on sail but not on spike.

aswaterman:
Good enough for me.

aswaterman:
Actually, good enough for me to believe the test cases are correct, but not good enough for being able to merge this :) We need it to be the case that executing `make run` in the `isa` directory succeeds.  I think that means you need to add `_zacas` to the ISA strings in `isa/Makefile` so that the extension will be enabled for Spike.  After doing that and making sure they run on Spike, we can merge.

ved-rivos:
Thanks. These and all other tests pass on spike but the breakpoint.S test seems to be failing. I am not sure if I caused it (I most likely did not since I only added amocas tests :) ). I will update the PR after that debug that to be sure.

ved-rivos:
Updated PR to extend the ISA string (breakpoint.S failure seems unrelated to addition of these tests).
FYI - SAIL PR: https://github.com/riscv/sail-riscv/pull/285

aswaterman:
Thanks, @ved-rivos. I’m sure you’re right, and I’ll look into the breakpoint test failure myself.

ved-rivos:
Found some time to look at this today. Seems like the test was not factoring in the case  that when `tcontrol` is not implemented, `mstatus.mie` must be 1 for trigger with `action=0` to match. I have sent PR #494 with a fix.

aswaterman:
@ved-rivos sorry we forgot to finish this up.  If it's fully working now, LMK.

PR #499, opened 2023-07-21
debug: Actually run tests in github workflow.

timsifive:
This should avoid problems like we just had where bad tests can break the OpenOCD workflow. These tests only run if any debug files are changed, so should have no impact at all on non-debug tests in this repo.


SiFiveGregS (approved):
with the caveat that I'm not experienced with the nitty gritty of github workflow, at a high level this all makes sense and looks good

PR #502, opened 2023-09-20
benchmarks: Various fixes for compiling with Clang

ekliptik:
#225 except -T behaves just fine with clang so we don't have to mess with that. Also renamed references to GCC to CC


aswaterman (commented):
Did you verify this all still works with GCC?

I think the replacement of `RISCV_GCC` with `RISCV_CC` is a little too ideological for a PR whose purpose is to fix things, but I guess it's OK.

ekliptik:
I can't address this immediately but I hope to play with riscv-tests some more next week and let you know

PR #503, opened 2023-09-26
Disable timer interrupt to fix some bugs

lz-bro:
Turn off the timer interrupt after the test case (such as InterruptTest) ends to avoid interference with other test cases.

### DIFF
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -1039,8 +1039,10 @@ class InterruptTest(GdbSingleHartTest):
             local = self.gdb.p("local")
             if interrupt_count > 1000 and \
                     local > 1000:
+                self.disable_timer()
                 return
 
+        self.disable_timer()
         assertGreater(interrupt_count, 1000)
         assertGreater(local, 1000)
 
@@ -1191,6 +1193,8 @@ class MulticoreRunAllHaltOne(GdbTest):
         time.sleep(1)
         self.gdb.p("buf", fmt="")
 
+        self.disable_timer(interrupt=True)
+
 class MulticoreRtosSwitchActiveHartTest(GdbTest):
     compile_args = ("programs/multicore.c", "-DMULTICORE")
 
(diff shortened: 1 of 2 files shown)


debug/gdbserver.py
```diff
@@ -1039,8 +1039,21 @@ def test(self):
             local = self.gdb.p("local")
             if interrupt_count > 1000 and \
                     local > 1000:
+                # Disable timer interrupt
+                for hart in self.target.harts:
+                    self.gdb.select_hart(hart)
+                    self.gdb.p("$mie")
+                    self.gdb.p("$mie=$mie & ~0x80")
+                    self.gdb.p("$mie")
```
timsifive, line 1047: Can you make this a helper function in GdbTest, and call that from the various places that need it?

You can probably remove the extra prints of $mie, too.

timsifive (changes requested):
I appreciate you're looking at these tests.

debug/gdbserver.py
lz-bro: I have made the changes as you suggested and conducted debug testing.

debug/gdbserver.py
```diff
@@ -1191,6 +1193,12 @@ def test(self):
         time.sleep(1)
         self.gdb.p("buf", fmt="")
 
+        # Disable timer interrupt
+        for hart in self.target.harts:
+            self.gdb.select_hart(hart)
+            self.gdb.interrupt()
+            self.gdb.p("$mie=$mie & ~0x80")
```
timsifive, line 1200: For more refactoring, can you make this a call to disable_timer() as well?

Probably you need a new `interrupt=False` argument to disable_timer(), and if it's True then it does the interrupt().

lz-bro: Your suggestions are very helpful and I have adopted it.

timsifive (approved):

PR #506, opened 2023-09-28
debug: Fix interrupt_all() to restore state.

timsifive:


debug/testlib.py
```diff
@@ -896,9 +896,10 @@ def interrupt(self, ops=None):
         return self.active_child.before.strip().decode()
 
     def interrupt_all(self):
-        for child in self.children:
-            self.select_child(child)
-            self.interrupt()
+        with PrivateState(self):
```
SiFiveGregS, line 899: OK, so active_child doesn't get clobbered as a side-effect, in the case where more than one GDB process is being managed

SiFiveGregS (approved):

debug/testlib.py
timsifive: Right. And this didn't actually cause any problem for me. I just thought it was, so I fixed it and it seems better to keep the fix.

PR #508, opened 2023-09-28
debug: Add Openocd.set_available()

timsifive:
This helper uses dmi_write commands to mark harts
available/unavailable.

### DIFF
--- a/debug/testlib.py
+++ b/debug/testlib.py
@@ -496,6 +496,30 @@ class Openocd:
             if time.time() - start > self.timeout:
                 raise TestLibError("Timed out waiting for targets to run.")
 
+    def set_available(self, harts):
+        """Set the given harts to available, and any others to be unavailable.
+        This uses a custom DMI register (0x1f) that is only implemented in
+        spike."""
+        available_mask = 0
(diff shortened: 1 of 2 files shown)


debug/testlib.py
```diff
@@ -496,6 +496,29 @@ def wait_until_running(self, harts):
             if time.time() - start > self.timeout:
                 raise TestLibError("Timed out waiting for targets to run.")
 
+    def set_available(self, harts):
+        """Set the given harts to available, and any others to be
+        unavailable."""
+        available_mask = 0
+        for hart in harts:
+            available_mask |= 1 << hart.id
+        self.command(f"riscv dmi_write 0x1f 0x{available_mask:x}")
```
SiFiveGregS, line 505: So, 0x1F in this context isn't a JTAG IR value (bypass), it's a DMI address ("custom"), which Spike implements "unavailable" control.

SiFiveGregS (approved):

debug/testlib.py
timsifive: Yes. I should probably add a comment explaining that.

PR #509, opened 2023-09-28
debug: Better interlock when interacting with gdb CLI.

timsifive:
Actually wait for the command to be echoed back. This means we won't be confused if there are extra newlines in gdb output.


SiFiveGregS (approved):
seems more robust approach

Issue #516, opened 2023-10-17
Does the IcountTest need to be improved?


lz-bro:
https://github.com/riscv-software-src/riscv-tests/blob/c36c814ae49610bc407a5c4e144866fa50e4510e/debug/gdbserver.py#L2059-L2083

In this test, count are set to 1 and 2 respectively. But, it is mentioned in RISC-V External Debug Support Version 0.13.2 that count may only support 1.
<img width="602" alt="icount" src="https://github.com/riscv-software-src/riscv-tests/assets/111737414/728f8660-4c17-4f0a-aeb6-7b714b196826">
If the hardware only supports count 1, but if count was set to 2, the hardware will process count to 1, causing set_trigger fail due to inconsistent writing and reading back of tdata1.

timsifive:
Most of these tests are written to run well against spike, and unfortunately not against more general hardware. I think it's great that you're going through and noting some of these issues.

What should happen on hardware that supports count=1 but not count=2 is that `riscv icount set m 2` returns an error.  OpenOCD will read back tdata1 after writing it, and should return an error because the hardware doesn't let count=2 be written.

From the test framework, this means that we need a hart property: max_icount or something. Then in the test we can confirm that either icount works, or we get an error when setting it.

lz-bro:
Thank you very much for your reply and suggestions.

PR #519, opened 2023-10-18
Support limits_icount_to_one in IcountTest

lz-bro:
https://github.com/riscv-software-src/riscv-tests/issues/516 This is taking into account that the hardware limits count to 1.

### DIFF
--- a/debug/targets.py
+++ b/debug/targets.py
@@ -138,6 +138,9 @@ class Target:
     # Supports controlling hart availability through DMCUSTOM.
     support_unavailable_control = False
 
+    # Instruction count limit
+    icount_limit = 4
+
     # Internal variables:
     directory = None
     temporary_files = []
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -2129,9 +2129,13 @@ class IcountTest(DebugTest):
     def test(self):
         # Execute 2 instructions.
         output = self.gdb.command("monitor riscv icount set m 2")
-        assertNotIn("Failed", output)
+        if self.target.icount_limit > 1:
+            assertNotIn("Failed", output)
+        else:
+            assertIn("Failed", output)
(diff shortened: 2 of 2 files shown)


lz-bro:
The UnavailableHaltedTest returned exception, I also tried `git reset --hard` https://github.com/riscv-software-src/riscv-tests/pull/515/commits/33663712bd8943fb360f4f5b730949ca11f3fc17 and run spike64-2 tests.

lz-bro:
My modifications don't seem to affect UnavailableTest, but UnavailableTest returns an exception in spike64-2 tests. I tried git reset --hard https://github.com/riscv-software-src/riscv-tests/commit/33663712bd8943fb360f4f5b730949ca11f3fc17 and ran the spike test locally, it returned an exception. However, I tried git revert https://github.com/riscv-software-src/riscv-tests/commit/33663712bd8943fb360f4f5b730949ca11f3fc17  and ran the spike test locally, it returned no exception. However, In commit https://github.com/riscv-software-src/riscv-tests/commit/33663712bd8943fb360f4f5b730949ca11f3fc17, the spike tests in github workflow return success,  I can't figure it out.

debug/targets.py
```diff
@@ -138,6 +138,9 @@ class Target:
     # Supports controlling hart availability through DMCUSTOM.
     support_unavailable_control = False
 
+    # When true it indicates that the hardware limits icount set to 1
+    limits_icount_to_one = False
+
```
timsifive, line 143: Instead of `limits_icount_to_one = False` how about `icount_limit = 4` (or whatever default seems reasonable)?

timsifive (changes requested):

debug/targets.py
lz-bro: I have changed it according to your opinion
.

timsifive (approved):

Issue #520, opened 2023-11-08
Some debug tests like are UnavailableHaltedTest broken


aap-sc:
@timsifive , @en-sc , @AnastasiyaChernikova  FYI.

The effects of the issue can be observed in  https://github.com/riscv/riscv-openocd/pull/928 (though this commit has nothing to do with the issue, the root cause seems to be in the test suite itself).

Basically, If I take current riscv-openocd TOT most **Unavailable...**  tests begin to fail in my environment. I've debugged the issue for a while now and I think I understand the root cause.

The problem is with `def command(self, cmd)` function from `debug/testlib.py`. What this function does is it tries to obtain results of OpenOCD command by parsing OpenOCD log instead of fetching the result directly (more on that later). Moreover, instead of sending only one command - we actually send 2 "commands". One real command and another is an artificial "command-like" comment to enforce OpenOCD to print prompt-like construct. The issue is that this approach does not always work.

Here are some examples (taken from log files after the failure).

**Example 1**

```
Debug: 5781 4300 riscv.c:3097 riscv_openocd_poll(): [riscv.cpu] should_remain_halted=0, should_resume=0
Debug: 5782 4300 target.c:3034 handle_target(): [riscv.cpu] target_poll() -> 0, next attempt in 100ms
targets
^@Debug: 5783 4313 command.c:152 script_debug(): command - targets
    TargetName         Type       Endian TapName            State       
--  ------------------ ---------- ------ ------------------ ------------
 0* riscv.cpu          riscv      little riscv.cpu          halted

> Debug: 5784 4399 riscv.c:3041 riscv_openocd_poll(): [riscv.cpu] Polling all harts.
Debug: 5785 4399 riscv.c:2864 riscv_poll_hart(): [riscv.cpu] polling, target->state=2
Debug: 5786 4399 riscv-013.c:393 scan(): 40b r 00000000 @11 -> + 00000000 @11; 0i
Debug: 5787 4399 riscv-013.c:403 scan(): dmstatus=0 { version=none, ...
Debug: 5788 4399 riscv-013.c:393 scan(): 40b - 00000000 @11 -> + 00430382 @11; 0i
Debug: 5789 4399 riscv-013.c:403 scan(): dmstatus=0 { version=none, ...}
Debug: 5790 4399 riscv.c:1664 riscv_flush_registers(): [riscv.cpu] Flushing register cache
[...]

PR #521, opened 2023-11-08
debug: Disable Unavailable tests.

timsifive:
They aren't always reliable. See #520.


timsifive:
@aap-sc, does that sound good? I assume we won't have this fixed today, and don't want to hold up other OpenOCD work.

aap-sc:
> @aap-sc, does that sound good? I assume we won't have this fixed today, and don't want to hold up other OpenOCD work.

Well... I've just created a patch with the fix. Took a while, though)

timsifive:
This problem was solved properly by #522.

PR #522, opened 2023-11-08
debug: use TCL-RPC to fetch results of OpenOCD commands instead of parsing log file

aap-sc:
Quick and dirty fix for https://github.com/riscv-software-src/riscv-tests/issues/520

### DIFF
--- a/debug/testlib.py
+++ b/debug/testlib.py
@@ -9,6 +9,7 @@ import tempfile
 import time
 import traceback
 
+import tty
 import pexpect
 import yaml
 
@@ -308,20 +309,16 @@ class Openocd:
             cmd = shlex.split(server_cmd)
         else:
             cmd = ["openocd"]
-            if debug:
-                cmd.append("-d")
 
         # This command needs to come before any config scripts on the command
         # line, since they are executed in order.
         cmd += [
             # Tell OpenOCD to bind gdb to an unused, ephemeral port.
             "--command", "gdb_port 0",
-            # We don't use the TCL server.
-            "--command", "tcl_port disabled",
-            # Put the regular command prompt in stdin. Don't listen on a port
-            # because it will conflict if multiple OpenOCD instances are running
-            # at the same time.
-            "--command", "telnet_port pipe",
+            # We create a socket for OpenOCD command line (TCL-RPC)
+            "--command", "tcl_port 0",
+            # don't use telnet
+            "--command", "telnet_port disabled",
         ]
 
         if config:
@@ -331,6 +328,7 @@ class Openocd:
                 sys.exit(1)
 
             cmd += ["-f", self.config_file]
+
         if debug:
             cmd.append("-d")
 
@@ -366,12 +364,18 @@ class Openocd:
         logfile.flush()
 
         self.gdb_ports = []
+        self.tclrpc_port = None
         self.start(cmd, logfile, extra_env)
+
+        self.openocd_cli = pexpect.spawn(f"nc localhost {self.tclrpc_port}")
+        # TCL-RPC uses \x1a as a watermark for end of message. We set raw
+        # pty mode to disable translation of \x1a to EOF
+        tty.setraw(self.openocd_cli.child_fd)
 
     def start(self, cmd, logfile, extra_env):
         combined_env = {**os.environ, **extra_env}
         # pylint: disable-next=consider-using-with
-        self.process = subprocess.Popen(cmd, stdin=subprocess.PIPE,
+        self.process = subprocess.Popen(cmd, stdin=None,
                 stdout=logfile, stderr=logfile, env=combined_env)
 
         try:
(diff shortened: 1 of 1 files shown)


aap-sc:
@timsifive could you please take a look?

aap-sc:
Note to reviewers:
1. The patch requires `nc` (netcat) utility to be present on a host system. I hope this is not a big deal since `nc` is pretty lightweight utility.
2.  It looks like I forgot to cleanup resources. If there is no objections against the suggested approach - I'll update the patch (I think I need to extend `__del__`  method of OpenOCD class to close the spawned process).

timsifive (approved):
This looks good to me. Using `nc` is a hack, but less of a hack than reading the logfile so it's an improvement.

I assume this fixes the failures you've been seeing.

debug/testlib.py
```diff
@@ -383,11 +387,16 @@ def start(self, cmd, logfile, extra_env):
...
-                if b"gdb" in m.group(1):
+                if b"gdb connections" in m.group(1):
                     self.gdb_ports.append(int(m.group(2)))
-                else:
+                elif b"tcl connections" in m.group(1):
+                    if self.tclrpc_port:
+                      raise TestLibError("unexpected re-definition of TCL-RPC port")
+                    self.tclrpc_port = int(m.group(3))
+
+                if self.tclrpc_port and (len(self.gdb_ports) > 0):
```
en-sc, line 399: The issue is not introduced by this patch, but I think it worth pointing out. This condition relies on all GDB ports being reported by OpenOCD prior to TCL-RPC port.
In turn this depends on whether the call to `init` is implicit or explicit:
```
> openocd ... -c "gdb_port 0" -c "tcl_port 0"  -c "telnet_port disabled" |& grep 'Listening'
Info : Listening on port 41587 for tcl connections
Info : Listening on port 40911 for gdb connections
Info : Listening on port 33639 for gdb connections
```
```
> openocd ... -c "gdb_port 0" -c "tcl_port 0"  -c "telnet_port disabled" -c init |& grep 'Listening'
Info : Listening on port 40575 for gdb connections
Info : Listening on port 38511 for gdb connections
Info : Listening on port 35989 for tcl connections
```
Currently, this works, since all the configs in `debug/targets/RISC-V` call `init` explicitly.
Wouldn't it be better to add some comment regarding this behavior?

debug/testlib.py
```diff
@@ -383,11 +387,16 @@ def start(self, cmd, logfile, extra_env):
...
+                        rb"Listening on port (\d+) for tcl connections)",
                         message="Waiting for OpenOCD to start up...")
-                if b"gdb" in m.group(1):
+                if b"gdb connections" in m.group(1):
                     self.gdb_ports.append(int(m.group(2)))
-                else:
+                elif b"tcl connections" in m.group(1):
+                    if self.tclrpc_port:
+                      raise TestLibError("unexpected re-definition of TCL-RPC port")
+                    self.tclrpc_port = int(m.group(3))
```
en-sc, line 397: I would suggest using named match groups and avoid searching a sub-string in a match:
```
                m = self.expect(
                        rb"Listening on port (?P<port>\d+) for (?P<server>(?:gdb)|(?:tcl)) connections",
                        message="Waiting for OpenOCD to start up...")
                if m["server"] == b"gdb":
                    self.gdb_ports.append(int(m["port"]))
                ...
```

en-sc: Moreover, since the regular expression has matched, `elif` can be replaced with `else` and an assertion, if you like.

debug/testlib.py
timsifive: Let's not add anything unnecessary to this patch. It's fairly urgent since without this fix, OpenOCD development is hamstrung. It's better to merge this soon than it is to disable the tests while we add more improvements to this PR.

timsifive:
Can you address the pylint complaints for this PR?

aap-sc:
> Can you address the pylint complaints for this PR

Sure! I'm in the middle of it...

debug/testlib.py
aap-sc: addressed. I haven't replaced `elif` with `else`, though

debug/testlib.py
aap-sc: I've already addressed this. Just added a comment

debug/testlib.py
```diff
@@ -422,18 +431,10 @@ def smp(self):
     def command(self, cmd):
         """Write the command to OpenOCD's stdin. Return the output of the
         command, minus the prompt."""
```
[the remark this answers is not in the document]
aap-sc: addressed

aap-sc:
> Can you address the pylint complaints for this PR?

Done.

Note: also addressed comments by @en-sc

timsifive:
pylint is still failing. (I'm commenting here because I'm not sure if you get notified about that automatically.)

aap-sc:
> pylint is still failing. (I'm commenting here because I'm not sure if you get notified about that automatically.)

My bad. I was in a hurry and submitted an update as soon as I've tested these changes with OpenOCD. Never bothered to check if a new issue is introduced. 

Addressed.

timsifive (approved):

en-sc (approved):

en-sc:
@timsifive, can you please re-run OpenOCD pipelines failed due to this issue:
https://github.com/riscv/riscv-openocd/actions/runs/6784073074/job/18460094331?pr=928
https://github.com/riscv/riscv-openocd/actions/runs/6822095851/job/18553551474?pr=959
https://github.com/riscv/riscv-openocd/actions/runs/6823520568/job/18557613980?pr=961

timsifive:
They're all running now.

PR #528, opened 2024-01-29
Uses appropriate addi instruction in lrsc test.

lvella:
How did the original version work in the first place?


aswaterman (approved):
@lvella `add` is an alias for `addi`, so it wasn't buggy before, but I don't have a problem with making the test a little more precise.

PR #537, opened 2024-02-04
Add sfence.vma in the disable_pmp function

lz-bro:
Change-Id: Ief659c0e68618f5b08a1eee98bc7df92817b8b51

### DIFF
--- a/debug/targets.py
+++ b/debug/targets.py
@@ -141,6 +141,10 @@ class Target:
     # Instruction count limit
     icount_limit = 4
 
+    # Implements page-based virtual memory. So when PMP changes execute an
+    # SFENCE.VMA
+    implements_page_virtual_memory = True
+
     # Support set_pmp_deny to create invalid addresses.
     support_set_pmp_deny = False
 
--- a/debug/testlib.py
+++ b/debug/testlib.py
@@ -1514,6 +1514,9 @@ class GdbTest(BaseTest):
                 # pmcfg0 readback matches write, so TOR is supported.
                 self.gdb.p("$pmpaddr0="
                            f"0x{(self.hart.ram + self.hart.ram_size) >> 2:x}")
+            if self.target.implements_page_virtual_memory:
+                # PMP changes require an sfence.vma, 0x12000073 is sfence.vma
+                self.gdb.command("monitor riscv exec_progbuf 0x12000073")
         except CouldNotFetch:
             # PMP registers are optional
             pass


lz-bro:
> SFENCE.VMA is illegal on implementations without S-mode. I think this should be conditionalized.

PMP changes require an SFENCE.VMA on any hart that implements page-based virtual memory, even if VM is not currently enabled. If the implementation without S-mode, does this mean that PMP registers are optional？

aswaterman:
As you said, "any hart that implements page-based virtual memory".  If S-mode is not implemented, then the hart doesn't implement page-based virtual memory.  So there's no contradiction here.

But I did misspeak earlier.  It is true that SFENCE.VMA is illegal on implementations without S-mode.  But it is also illegal on implementations _with_ S-mode but without virtual memory.  So, it needs to be conditionalized on "is S-mode implemented" _and_ "is page-based virtual memory implemented".  You can determine the latter property by checking whether `satp.MODE` is hardwired to 0.

But note that writing `satp` is illegal unless S-mode is implemented.  So you should first check if S-mode is implemented, then check if `satp.MODE` is writable, and only then should you execute SFENCE.VMA.

lz-bro:
> As you said, "any hart that implements page-based virtual memory" needs to execute SFENCE.VMA. If S-mode is not implemented, then the hart doesn't implement page-based virtual memory. So there's no contradiction here.
> 
> But I did misspeak earlier. It is true that SFENCE.VMA is illegal on implementations without S-mode. But it is also illegal on implementations _with_ S-mode but without virtual memory. So, it needs to be conditionalized on "is S-mode implemented" _and_ "is page-based virtual memory implemented". You can determine the latter property by checking whether `satp.MODE` is hardwired to 0.
> 
> But note that writing `satp` is illegal unless S-mode is implemented. So you should first check if S-mode is implemented, then check if `satp.MODE` is writable, and only then should you execute SFENCE.VMA.

Thank you for your explanation，I've introduced an implements_page_virtual_memory property for the target. But, There is a pipeline failure in Build Spike.

aswaterman:
I think we need to do something similar to the following.  Can you give it a try and see if you can make it work?  https://github.com/riscv-software-src/riscv-isa-sim/pull/1584/files

lz-bro:
> I think we need to do something similar to the following. Can you give it a try and see if you can make it work? https://github.com/riscv-software-src/riscv-isa-sim/pull/1584/files

Yes, It works now.

lz-bro:
@aswaterman Would you kindly review and merge it.

aswaterman (approved):

PR #539, opened 2024-02-17
Add bitmanip test cases

rogerchang23424:
The test cases contains zba, zbb, zbc, zbs instructions


isa/rv32ub/Makefrag
```diff
@@ -0,0 +1,36 @@
+#=======================================================================
+# Makefrag for rv64ub tests
+#-----------------------------------------------------------------------
+
+rv32ub_sc_tests = \
+	andn \
+	bclr bclri \
+	bext bexti \
+  binv binvi \
```
aswaterman, line 9: tabs in this file, not spaces

aswaterman (commented):
Can you separate the tests into directories per extension, rather than lumping them together into "b"?

Can you tell us how you tested these?  Unfortunately, I have an older version of RISC-V binutils on this machine, so I can't compile and run the tests on Spike myself.

isa/rv32ub/Makefrag
rogerchang23424: Done

rogerchang23424:
> Can you separate the tests into directories per extension, rather than lumping them together into "b"?

Done

> Can you tell us how you tested these? Unfortunately, I have an older version of RISC-V binutils on this machine, so I can't compile and run the tests on Spike myself.

The test cases require RISC-V gcc 12+ and binutils 2.38.0+ to run them, and make sure the version of Spike is latest.

https://godbolt.org/z/ehaE8Whb4

aswaterman (approved):

aswaterman:
Thank you for this helpful contribution!

PR #540, opened 2024-03-01
Add hypervisor support for PrivRw

lz-bro:
Change-Id: I635a8f86b407f104ea74f00f4aefa1ce8170cb2d

### DIFF
--- a/debug/gdbserver.py
+++ b/debug/gdbserver.py
@@ -1668,14 +1668,15 @@ class PrivTest(GdbSingleHartTest):
         # pylint: disable=attribute-defined-outside-init
         self.gdb.load()
 
-        misa = self.hart.misa
         self.supported = set()
-        if misa & (1<<20):
+        if self.hart.extensionSupported("U"):
             self.supported.add(0)
-        if misa & (1<<18):
+        if self.hart.extensionSupported("S"):
             self.supported.add(1)
-        if misa & (1<<7):
-            self.supported.add(2)
+        if self.hart.extensionSupported("H"):
+            self.supported_vmodes = set(self.supported)
+            for prv in self.supported_vmodes:
+                self.supported.add((1 << 2) + prv)
         self.supported.add(3)
 
         self.disable_pmp()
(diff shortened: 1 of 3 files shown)


lz-bro:
@aswaterman Would you kindly take a look?

aswaterman (commented):
I'm not familiar with this code, so I'd rather @aap-sc, @en-sc, or @timsifive sign off on this.

debug/gdbserver.py
```diff
@@ -1649,14 +1649,16 @@ def setup(self):
         # pylint: disable=attribute-defined-outside-init
         self.gdb.load()
 
-        misa = self.hart.misa
+        # It may be affected by reset halt
+        misa = self.gdb.p(f"$misa=0x{self.hart.misa:x}")
```
en-sc, line 1653: I don't quite get it. Which part is responsible for setting `misa.H`?

debug/gdbserver.py
```diff
@@ -1649,14 +1649,16 @@ def setup(self):
...
         self.supported = set()
         if misa & (1<<20):
             self.supported.add(0)
         if misa & (1<<18):
             self.supported.add(1)
         if misa & (1<<7):
-            self.supported.add(2)
+            self.supported.add(4)
+            self.supported.add(5)
```
en-sc, line 1661: Moreover, AFAIU, `H` extension does not require presence of both `S` and `U` extensions.
Therefore, I'd suggest something like:
```suggestion
            self.supported_vmodes = self.supported
        else:
            self.supported_vmodes = None
```

debug/gdbserver.py
lz-bro: When the power is turned on, the hart starts executing the brom code.

lz-bro: When the power is turned on, the hart starts executing the brom code. Here does not affect this case, repeated testing will affect CheckMisa.

debug/gdbserver.py
lz-bro: Thanks, addressed

lz-bro:
Ping

en-sc (approved):
LGTM. Sorry for the delay.

aswaterman (approved):

PR #541, opened 2024-03-01
[debug] Support mcontrol and mcontrol6 based on HwbpManual

lz-bro:
Change-Id: I2ee8718f03d6e381369b7d3bb0222563789ed525


debug/gdbserver.py
```diff
@@ -710,6 +715,7 @@ def test(self):
             self.gdb.p("$pc")
             assertRegex(output, r"[bB]reakpoint")
             assertIn("rot13 ", output)
+            self.gdb.stepi()
```
aap-sc, line 718: why do we need this `stepi` ?

debug/gdbserver.py
```diff
@@ -697,8 +699,11 @@ def test(self):
             if value != tselect:
                 raise TestNotApplicable
             self.gdb.p(f"$tdata1=0x{tdata1:x}")
-            value = self.gdb.p("$tselect")
-            if value == tdata1:
+            value = self.gdb.p("$tdata1")
+            if value == tdata1 or \
+                    (value >> (xlen-4) == 0x6 and \
```
aap-sc, line 704: `value >> (xlen-4) == 0x6` What is the purpose of this condition? If I read this right we explicitly ask for TYPE=2 trigger. How come we end up with TYPE=6?

debug/gdbserver.py
lz-bro: https://github.com/riscv-software-src/riscv-tests/blob/7dfa49bb33ac2393742232cb6f77b79ec5c9df47/debug/programs/debug.c#L65-L68
According to the program, I think that the breakpoint should be hit 2 times at rot13, and then hit at crc32a and _exit sequentially. So stepi avoids hitting the rot13 breakpoint at the same position as main

debug/gdbserver.py
lz-bro: The purpose is to be compatible with spec 1.0 and 0.13. On our hardware that supports spec 0.13, setting the TYPE to 2  trigger actually we end up with TYPE=6.

aap-sc: can we instead just re-request type==6 if the resulting value of tdata1 is different? I believe this would enable us to support wide range of platforms in a more cleaner way.

debug/gdbserver.py
aap-sc: > So stepi avoids hitting the rot13 breakpoint at the same position as main

I don't get it. What do you mean "at the same position as main" ?  main and rot13 should have different addresses. When gdb issues a continue request it issues a single-step and then issues a real continue. We should breakpoint at main should not be hit.

lz-bro:
@aap-sc hello，I have made some modifications to support wide range of platforms. Could you please review it.

aap-sc:
@lz-bro I think it would be better to leave `set_manual_trigger` and just perform 2 separate calls (for type2 and type6) passing the desired mask along the way (if type2/type6 triggers require a different mask). @en-sc would you kindly take a look too?

lz-bro:
> @lz-bro I think it would be better to leave `set_manual_trigger` and just perform 2 separate calls (for type2 and type6) passing the desired mask along the way (if type2/type6 triggers require a different mask). @en-sc would you kindly take a look too?

Thanks, I agree that perform 2 separate calls (for type2 and type6)

debug/gdbserver.py
```diff
@@ -717,21 +736,27 @@ def set_manual_trigger(self, tdata1, tdata2):
...
-            type_none = set_field(0, MCONTROL_TYPE(self.hart.xlen),
-                                  MCONTROL_TYPE_NONE)
-            if type_rb == type_none:
+            type_rb = tdata1_rb & TDATA1_TYPE(self.hart.xlen)
+            type_match = set_field(0, TDATA1_TYPE(self.hart.xlen),
+                                   TDATA1_TYPE_MATCH)
+            type_none = set_field(0, TDATA1_TYPE(self.hart.xlen),
+                                  TDATA1_TYPE_NONE)
+            if type_rb == type_match:
+                maskmax_rb = tdata1_rb & MCONTROL_MASKMAX(self.hart.xlen)
```
en-sc, line 746: I'd sugest passing the mask to use before the comparison of `tdata1` and `tdata1_rb` as an argument of `set_manual_trigger()`

lz-bro:
@en-sc Passing both type_none and type_match as an argument of set_manual_trigger() ？Which parameter are you referring to as a function argument

lz-bro:
Ping

en-sc (approved):
LGTM. Sorry for the delay

aswaterman (approved):

PR #542, opened 2024-03-01
fix CI for OpenOCD/debugger tests

aap-sc:
`git ls-remote` lists both `refs/head` and `refs/remotes`. riscv-openocd got
a new entry in refs/remotes that matches `riscv` pattern around Jan 25.
This breaks gitlab pipeline. The commit restricts ls-remote only to
`refs/head`.

In addition this commit changes the address of riscv-openocd repository,
since it was moved to riscv-collab space a while ago.

### DIFF
--- a/.github/workflows/spike-openocd-tests.yml
+++ b/.github/workflows/spike-openocd-tests.yml
@@ -7,7 +7,7 @@ env:
   SPIKE_REV: master
   RISCV_TESTS_REPO: https://github.com/riscv-software-src/riscv-tests.git
   RISCV_TESTS_REV: master
-  OPENOCD_REPO: https://github.com/riscv/riscv-openocd.git
+  OPENOCD_REPO: https://github.com/riscv-collab/riscv-openocd.git
   OPENOCD_REV: riscv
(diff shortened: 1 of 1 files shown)


aap-sc:
This PR should fix the issue: https://github.com/riscv-software-src/riscv-tests/pull/542

aap-sc:
@lz-bro , @aswaterman FYI.

@en-sc could you please review this one?

en-sc:
Seems like the other match, which cased trouble (`refs/remotes/riscv-openocd/riscv`) was added to `riscv-openocd` by mistake. I have deleted it and the pipelines should work without the patch from now on ([re-run example](https://github.com/riscv-software-src/riscv-tests/actions/runs/7773075506)).

However, the link to `riscv-openocd` repo should still be updated.

As for the part regarding revisions, I do think we can use a more robust mechanism here.
I see the following alternatives:
* Change `OPENOCD_REV` to `HEAD` -- this will protect against changing the default branch in riscv-openocd
* Change `OPENOCD_REV` to `refs/tags/latest` -- this is the tag, updated by [`snapshot.yml`](https://github.com/riscv-collab/riscv-openocd/blob/riscv/.github/workflows/snapshot.yml) workflow on push. IMHO, this is the best option, since this tag is the closest thing to a release in `riscv-openocd`.
* Leave it as-is, since if it works -- don't touch it :)

aap-sc:
> I have deleted it and the pipelines should work without the patch from now on ([re-run example]

Yep. It works now. I'll abandon the patch then

PR #547, opened 2024-03-19
rv64i: ma_addr can check mis-aligned or access exception

chihminchao:
For illegal mis-aligned memory access, the spec allows implementation to trigger mis-aligned fault or access fault. The origin case only check mis-aligned fault and the patch adds access-fault to the end of check.

### DIFF
--- a/isa/rv64mi/ma_addr.S
+++ b/isa/rv64mi/ma_addr.S
@@ -96,7 +96,9 @@ RVTEST_CODE_BEGIN
   .global mtvec_handler
 mtvec_handler:
   csrr t0, mcause
-  bne t0, s1, fail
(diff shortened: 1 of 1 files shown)


rakeshpalak (approved):

aswaterman:
I agree we need to do something like this, but this version is too permissive: it allows any misaligned access to report a load access fault or a store access fault, regardless of whether it was actually a load or a store.  I tried my  hand at an alternate version, based upon yours.  Can you confirm it works for you, and close the PR if so?  fd01e46314636666b5bed1f08465f734751a4cda

chihminchao:
@aswaterman Your solution works and is better that ours.

aswaterman:
Cool, thanks for checking.

PR #553, opened 2024-05-01
[debug tests] increase remotetimeout for all spike-based targets

aap-sc:
Spike simulator is very demanding to CPU resources. This causes debug tests to sporadically fail on slower machines. Increasing of gdb's `remotetimeout` should get rid of such failures, unless we run the testsuite on a potato.

The only downside is that if OpenOCD is broken, tests can run longer. However, I think this is the sacrifice we can make, since execution time is not affected if everything works as expected.


aap-sc:
@en-sc , please, take a look.

@TommyMurphyTM1234 FYI

TommyMurphyTM1234:
For reference purposes:

* https://github.com/riscv-collab/riscv-openocd/issues/1049#issuecomment-2088496775

aswaterman (approved):
Seems fine to me.

PR #562, opened 2024-06-10
Add support for Zalasr unratified extension

mehnadnerd:
Based on https://github.com/riscv-software-src/riscv-tests/pull/492.
I was able to test it with my hacked-up clang/spike and makefiles, so I don't think this can be merged until my patches for them are merged.
Specification is available at https://github.com/riscv/riscv-zalasr


aswaterman:
Thanks.  Agreed we need to wait to merge it.

Issue #546, opened 2024-03-08
 `--with-xlen=32` builds are failing


bluewww:
In `riscv-tests/benchmarks/Makefile`, the ABI is [hardcoded](https://github.com/riscv-software-src/riscv-tests/blob/master/benchmarks/Makefile#L46) to `lp64d`.

This causes  `XLEN=32` builds to fail like this

```bash
mkdir -p benchmarks
make -C benchmarks -f /tmp/riscv-tests/benchmarks/Makefile src_dir=/tmp/riscv-tests/benchmarks XLEN=32 
make[1]: Entering directory '/tmp/riscv-tests/benchmarks'
[...]

PR #566, opened 2024-06-13
Benchmarks: parameterise ABI based on XLEN

TommyMurphyTM1234:
Addresses this issue:

- https://github.com/riscv-software-src/riscv-tests/issues/546

Parameterise `ABI` based on selected `XLEN` unless it has already been passed in.

Note that since `RISCV_PREFIX` has always been parameterised using `XLEN`, this means that `XLEN == 32` expects, by default, to use a `riscv32-unknown-elf-` prefixed toolchain. If necessary, this can be overridden as follows:
```
./configure ... --with-xlen=32
make benchmarks RISCV_PREFIX=riscv64-unknown-elf-
```
Obviously, whatever toolchain prefix is used the toolchain needs to support the selected `arch/abi` - e.g. `rv64gcv/lp64d` or `rv32gcv/ilp32d` or whatever other `ABI` is passed in - via the toolchain's default `arch/abi` or via multilibs for the selected `arch/abi`.

I didn't consider `XLEN == 128` since, as far as I can see, it isn't considered elswhere in the test suite but it would be trivial to extend this mod to cover it if necessary.

### DIFF
--- a/benchmarks/Makefile
+++ b/benchmarks/Makefile
@@ -6,6 +6,13 @@
 
 XLEN ?= 64
 
+ifeq ($(XLEN),32)
+ABI ?= ilp32d
+endif
+ifeq ($(XLEN),64)
+ABI ?= lp64d
+endif
+
 default: all
 
 src_dir = .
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

TommyMurphyTM1234:
Thanks @aswaterman. 👍

aswaterman:
You’re welcome!

PR #567, opened 2024-06-20
Use Zvl/Zve to communicate VLEN/ELEN to target in debug tests

jerryz123:
Goes with https://github.com/riscv-software-src/riscv-isa-sim/pull/1701

PR #574, opened 2024-08-07
Remove `debug/test` binary file

en-sc:
The file was added by 00ab5f0dd4cf56b5a0551bc5adedf60c765d0c66 in #567. Seems like a mistake.
```
> readelf -h debug/test
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              EXEC (Executable file)
  Machine:                           RISC-V
  Version:                           0x1
  Entry point address:               0x1212340000
  Start of program headers:          64 (bytes into file)
  Start of section headers:          16552 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         2
  Size of section headers:           64 (bytes)
  Number of section headers:         18
  Section header string table index: 17
```


jerryz123 (approved):
Oops, sorry

aswaterman (approved):
Whoops, I should’ve noticed that in review.

PR #575, opened 2024-08-07
[debug] Reserve the trigger in `HwbpManual`

en-sc:
After https://github.com/riscv-collab/riscv-openocd/pull/1111 is merged, the registers a user wishes to have direct control of should be reserved.
This is the case in `HwbpManual`.

The test still works with older OpenOCD versions, since no exception is generated when a command (`riscv reserve_trigger` in this case) is not found.


aap-sc (approved):
1. LGTM, though we should still wait till the MR is merged.
2. I think we'd better to modify testsuite to throw errors if command results in failure.

PR #579, opened 2024-08-26
Fix EtriggerTest assuming that NULL causes a trap

Wren6991:
Currently the test is broken on systems which map address 0 normally. There is already a mechanism for the test target to supply a known-bad address, so use that address if it is provided.


aap-sc (approved):

aswaterman (approved):
This strikes me as brittle, but I think it’s fine as a workaround.

Issue #584, opened 2024-10-18
Compilation issue for dhrystone benchmark


pratikkedar:
dhrystone benchmark results in compilation issues.

/home/test/code-arena/riscv-tests/benchmarks/dhrystone/dhrystone.c:20:1: error: return type defaults to 'int' [-Wimplicit-int]
   20 | Proc_6 (Enum_Val_Par, Enum_Ref_Par)
      | ^~~~~~
/home/test/code-arena/riscv-tests/benchmarks/dhrystone/dhrystone.c: In function 'Proc_6':
/home/test/code-arena/riscv-tests/benchmarks/dhrystone/dhrystone.c:29:9: error: implicit declaration of function 'Func_3' [-Wimplicit-function-declaration]
   29 |   if (! Func_3 (Enum_Val_Par))
      |         ^~~~~~
/home/test/code-arena/riscv-tests/benchmarks/dhrystone/dhrystone.c: At top level:
/home/test/code-arena/riscv-tests/benchmarks/dhrystone/dhrystone.c:54:1: error: return type defaults to 'int' [-Wimplicit-int]
   54 | Proc_7 (Int_1_Par_Val, Int_2_Par_Val, Int_Par_Ref)
      | ^~~~~~
/home/test/code-arena/riscv-tests/benchmarks/dhrystone/dhrystone.c:74:1: error: return type defaults to 'int' [-Wimplicit-int]
   74 | Proc_8 (Arr_1_Par_Ref, Arr_2_Par_Ref, Int_1_Par_Val, Int_2_Par_Val)
      | ^~~~~~
/home/test/code-arena/riscv-tests/benchmarks/dhrystone/dhrystone_main.c: In function 'main':
/home/test/code-arena/riscv-tests/benchmarks/dhrystone/dhrystone_main.c:120:7: error: implicit declaration of function 'Proc_5' [-Wimplicit-function-declaration]
  120 |       Proc_5();
      |       ^~~~~~
/home/test/code-arena/riscv-tests/benchmarks/dhrystone/dhrystone_main.c:121:7: error: implicit declaration of function 'Proc_4' [-Wimplicit-function-declaration]
  121 |       Proc_4();
      |       ^~~~~~
/home/test/code-arena/riscv-tests/benchmarks/dhrystone/dhrystone_main.c:127:21: error: implicit declaration of function 'Func_2'; did you mean 'Func_1'? [-Wimplicit-function-declaration]
[...]

PR #585, opened 2024-10-18
fixing compilation issue for dhrystone benchmark

pratikkedar:
Compilation issue for dhrystone benchmark #584


aswaterman (changes requested):
Dhrystone is strict about not changing the source code.  The right fix is to change the compiler options so that the code becomes legal again.

PR #606, opened 2025-04-07
CI: Add a CI to build ELF

Arielfoever:
just for some... You can use it in https://github.com/riscv/sail-riscv/tree/master/test/riscv-tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,87 @@
+# Build Spike and run a couple of debug tests.
+
+name: Build ELF
+
+env:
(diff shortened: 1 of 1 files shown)


trdthg:
@aswaterman Any suggestions? 

Our starting point is that we want to integrate riscv-tests tests into sail-riscv. Our options are as follows:

- download test binary from ci release
	- This is what this PR does. Downloading test files in CI maybe is not a good idea. Although we could cache it in CI.
- using riscv-tests as submodule, and build it in ci
	- I personally think this method is the best, although I don't think git submodule is particularly easy to use.
- build or download binary and keep it in reposity. 
	- Easiest method for users, they don't need to configure riscv toolchain or anything else locally, but the disadvantage is also obvious, it takes up a lot of disk space

Anyway, I think there is no harm in distributing binaries in this repository.

PR #608, opened 2025-04-07
Readme: Add autoupdate.

Arielfoever:
Avoid

```
configure.ac:14: warning: AC_OUTPUT should be used without arguments.
configure.ac:14: You should run autoupdate.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ https://github.com/riscv-collab/riscv-gnu-toolchain) package is installed.
     $ git clone https://github.com/riscv/riscv-tests
     $ cd riscv-tests
     $ git submodule update --init --recursive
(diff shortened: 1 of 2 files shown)


aswaterman:
I'd rather just commit the updated `configure.ac` and potentially `configure` (though it doesn't appear this change to `configure.ac` actually changes `configure` in this case).

Arielfoever:
> I'd rather just commit the updated `configure.ac` and potentially the updated `configure` (though it doesn't appear this change to `configure.ac` actually changes `configure` in this case).

As you wish.

aswaterman:
Manually merged.  (I cherry-picked the updated `configure.ac` and not the README change)

PR #609, opened 2025-04-10
Add test for pmpaddr[G-1] edge cases

Timmmm:
This adds a test for pmpaddr[G-1] edge cases. This bit is unusual because it has read legalisation that depends on another CSR.

If the pmpcfg mode is OFF/TOR then it reads as 0. Otherwise it reads as its underlying value. Irrespective of pmpcfg it is always writable so you can always write it even when it reads as zero.

Another edge case is when CSRS/C are used to modify a *different* bit and the underlying value of pmpaddr[G-1] is 1 but it reads as 0. The spec says "All CSR instructions atomically read-modify-write a single CSR" so even if these instructions don't directly modify pmpaddr[G-1] they do read it and then write it back, which will change its underlying value since it reads as 0.

Tested against the Sail model with G=0,1,2,10,30. To test:

1. Build the Sail emulator as described [here](https://github.com/riscv/sail-riscv?tab=readme-ov-file#building-the-model).
2. Edit `config/default.json` and change `grain` to 10 or whatever.
3. Run `build/c_emulator/riscv_sim_rv64d` --config config/default.json /path/to/riscv-tests/isa/rv64mi-p-pmpaddr`

### DIFF
--- a/isa/rv32mi/Makefrag
+++ b/isa/rv32mi/Makefrag
@@ -18,5 +18,6 @@ rv32mi_sc_tests = \
 	sw-misaligned \
 	zicntr \
 	instret_overflow \
+	pmpaddr \
 
 rv32mi_p_tests = $(addprefix rv32mi-p-, $(rv32mi_sc_tests))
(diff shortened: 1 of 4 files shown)


aswaterman (approved):
While I confess I didn't work through all cases, it passes the smell test, and it's a helpful contribution.  Thanks, @Timmmm!

PR #617, opened 2025-04-27
[debug] check triggers exist in icount/itrigger/etrigger test

lz-bro:


debug/testlib.py
```diff
@@ -1545,6 +1545,19 @@ def classSetup(self):
         GdbTest.classSetup(self)
         self.parkOtherHarts()
 
+    def check_triggers_exist(self, type_bitmask):
```
en-sc, line 1548: I'd suggest avoiding marking a test as unsupported based on run-time information as much as possible.
My suggestion is:
1. Introduce a new hart property: `supports_etrigger`, etc. By default it is set to `True` and can be redefined in a fashion similar to other target properties:
https://github.com/riscv-software-src/riscv-tests/blob/53936b89ae8dc4e9d91120bec0e43c1c6263da17/debug/targets/RISC-V/spike32.py#L18
2. In the test, test that the target description is correct, meaning:
  - if `target supports_etrigger` is true the command functions as expected.
  - else the command returns an error.

en-sc (changes requested):

debug/testlib.py
lz-bro: Addressed

en-sc: Please, add a `TODO` that there should be a check that a trigger is really not supported if it is marked as unsupported.

en-sc (approved):

lz-bro:
@aswaterman can we merge this?

aswaterman (approved):
Yep.

lz-bro:
@aswaterman Done.

aswaterman:
Thanks, I’ll go ahead and merge.

PR #618, opened 2025-04-28
CI: Bump to Ubuntu latest

lz-bro:
#617  has a pipeline failure for Ubuntu 20.04 has retired.

### DIFF
--- a/.github/workflows/spike-openocd-tests.yml
+++ b/.github/workflows/spike-openocd-tests.yml
@@ -34,16 +34,16 @@ on:
 jobs:
   test:
     name: Test debug (Ubuntu)
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: recursive
 
       - name: Install packages
         run: |
           sudo apt-get update
-          sudo apt-get install -y device-tree-compiler build-essential
+          sudo apt-get install -y device-tree-compiler build-essential libjim-dev
 
       - name: Get revisions of dependencies
         run: |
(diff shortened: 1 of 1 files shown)


en-sc (approved):
LGTM

.github/workflows/spike-openocd-tests.yml
```diff
@@ -34,16 +34,16 @@ on:
 jobs:
   test:
     name: Test debug (Ubuntu)
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
en-sc, line 37: Alternatively, Ubuntu 24.04 can be used. However, AFAIU, GitHub runner supports the two latest LTS versions of Ubuntu, so I'm also leaning towards using `ubuntu-latest`.
FIY: https://github.com/actions/runner-images/issues/11101

en-sc, line 37: Alternatively, Ubuntu 24.04 can be used. However, AFAIU, GitHub runner supports the two latest LTS versions of Ubuntu, so I'm also leaning towards using `ubuntu-latest`.
FIY: https://github.com/actions/runner-images/issues/11101

.github/workflows/spike-openocd-tests.yml
```diff
@@ -34,16 +34,16 @@ on:
...
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: recursive
 
       - name: Install packages
         run: |
           sudo apt-get update
-          sudo apt-get install -y device-tree-compiler build-essential
+          sudo apt-get install -y device-tree-compiler build-essential libjim-dev
```
en-sc, line 46: Just a clarification:
`libjim-dev` is the new requirement of OpenOCD. Previously it was build in-tree.

en-sc:
@aswaterman, can you please take a look at this one?
This is quite urgent, since the CI is broken right now.

aswaterman:
@en-sc Yes, thank you for the ping.

I don't like using `ubuntu-latest` because we get sudden breakages when github decides to redefine "latest".  I'll upgrade to 24.04 instead.

aswaterman:
Closed in favor of #619

PR #634, opened 2025-11-16
Clear mcountinhibit.ir in minstret test to ensure minstret increments.

Wren6991:
This test assumes that `minstret` is already running, but the spec does not mandate a reset value for `mcountinhibit.`

My implementation deliberately resets `mcountinhibit` to all-ones to save energy and I expect this is fairly common.

### DIFF
--- a/isa/rv64mi/instret_overflow.S
+++ b/isa/rv64mi/instret_overflow.S
@@ -13,6 +13,15 @@
 RVTEST_RV64M
 RVTEST_CODE_BEGIN
 
+  # Clear mcountinhibit.ir to ensure the counter runs (reset value is unspecified).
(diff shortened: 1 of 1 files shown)


aswaterman (commented):
`mcountinhibit` is an optional CSR; this change will make the test fail on implementations that don't have it.  You can do something like the following to benignly handle the illegal-instruction exception:

```
  la t0, 1f
  csrrw t0, mtvec, t0
  csrci mcountinhibit, 0x4
  .balign 4
1:
  csrw mtval, t0
```

Wren6991:
Thanks, I'd completely missed that the optionality of `mcountinhibit` was independent of the actual counters.

I applied the suggested change.

PR #636, opened 2025-11-21
debug: test implements_page_virtual_memory before executing sfence.vma instruction

hirooih:
This PR is separate from #632.

As discussed in  #537, `sfence.vma` instruction requires S-mode. This test is required as `disable_pmp()` is.


aswaterman:
@en-sc can you take a look at this one, too

PR #638, opened 2026-01-05
syscalls: use correct printf code for counters[]

xypron:
Building benchmarks/common/syscalls.c results in a warning:

    warning: format ‘%d’ expects argument of type ‘int’, but argument 4
    has type ‘uintptr_t’ {aka ‘long unsigned int’} [-Wformat=]
      118 |       pbuf += sprintf(pbuf, "%s = %d\n", counter_names[i], counters[i]);
          |                                   ~^                       ~~~~~~~~~~~
          |                                    |                               |
          |                                    int                             uintptr_t {aka long unsigned int}

Use %lu for to print uintptr_t.


xypron:
Thanks for reviewing @aswaterman I have changed the patch to use %lu.

PR #640, opened 2026-01-07
benchmarks: undefined reference to `__printf_chk'

xypron:
When building with Ubuntu's gcc 15.2.0-7ubuntu1 an error occurs:

    undefined reference to `__printf_chk'

This is due to _FORTIFY_SOURCE being set by the toolchain defaults.

Undefine the symbol.


aswaterman:
Didn't know about this one; makes sense.

PR #641, opened 2026-01-07
Avoid RWX section

xypron:
Building leads to a warning

    riscv64-linux-gnu/bin/ld: warning:
    median.riscv has a LOAD segment with RWX permissions

The text sections should we RX only.


aswaterman (approved):
Thanks for fixing this one; it has been annoying me but I never got around to it.

PR #643, opened 2026-01-07
Make build compatible with distro tooling

xypron:
On RISC-V Linux distros like Debian or Ubuntu typically only a 64bit toolchain is provided
using the lp64d ABI.

The first patch avoids building 32bit artifacts if XLEN is 64 bit.

The second patch makes use of default 64 bit ABI which is lp64d.


aswaterman (changes requested):
64-bit toolchains can compile 32-bit tests.  What problem is this solving?

PR #647, opened 2026-01-23
add tests for the SPMP extension

ybc-alkaid:
This PR introduces basic tests for the RISC-V S-level Physical Memory Protection (SPMP) extension.

Here is the pointer to the SPMP specification:
https://github.com/riscv/riscv-isa-manual/pull/2573

Here is a pre-built PDF of SPMP:
https://github.com/riscv/riscv-spmp/blob/main/rv-spmp-spec.pdf


benchmarks/Makefile
```diff
@@ -39,6 +39,7 @@ base_bmarks = \
 	mt-matmul \
 	mt-memcpy \
 	pmp \
+	spmp \
```
aswaterman: Seems like this is a mistake?

aswaterman (commented):
Thanks, @ybc-alkaid.  I see these are not included in `isa/Makefile`, presumably because they don't yet work on upstream Spike.

My preference is to add them to `isa/Makefile`, but wait to merge this PR until Spike can execute them.

benchmarks/Makefile
ybc-alkaid: Sorry about that, I'll fix it.

ybc-alkaid:
Hi @aswaterman,
The tests have been added to isa/rv64si/Makefrag and isa/rv32si/Makefrag, so they will be included in isa/Makefile. It this the right way?

aswaterman:
@ybc-alkaid My mistake, it looks like you did it correctly.

How did you test these tests?  Do you have a prototype of SPMP in Spike so that you can successfully execute `make run` in the `isa/` directory?

ybc-alkaid:
@aswaterman Yes, I test it via my Spike prototype: https://github.com/ybc-alkaid/riscv-isa-sim
This implementation is only for basic tests, and does not support resource sharing (i.e., mpmpdeleg).
The simulator that complies with the SPMP specification is QEMU: https://github.com/luisccc/qemu/tree/feat/spmp

> execute make run in the isa/ directory

That requires a minor modification to the Makefile. I've updated it.

isa/Makefile
```diff
@@ -74,10 +70,17 @@ vpath %.S $(src_dir)
...
 %.out32: %
-	$(RISCV_SIM) --isa=rv32gc_ziccid_zfh_zicboz_svnapot_zicntr_zba_zbb_zbkb_zbkx_zbc_zbs_zicclsm $< 2> $@
+	$(RISCV_SIM) --isa=rv32gc_ziccid_zfh_zicboz_svnapot_zicntr_zba_zbb_zbc_zbs_zicclsm $< 2> $@
+
+# SPMP tests need spmp extension
+rv64si-p-spmpaddr.out rv64si-p-spmpcfg.out: %.out: %
+	$(RISCV_SIM) --isa=rv64gch_ziccid_zfh_zicboz_svnapot_zicntr_zba_zbb_zbc_zbs_zicclsm_sspmp $< 2> $@
+
+rv32si-p-spmpaddr.out32 rv32si-p-spmpcfg.out32: %.out32: %
+	$(RISCV_SIM) --isa=rv32gc_ziccid_zfh_zicboz_svnapot_zicntr_zba_zbb_zbc_zbs_zicclsm_sspmp $< 2> $@
```
aswaterman: Normally, we unconditionally enable the extension (even though most tests don't need it).  Can we do that here, too?  (I understand that it won't work until the Spike changes are merged.)

aswaterman:
> Yes, I test it via my Spike prototype

Cool.  Once the Spike stuff is merged, we can merge this PR, too.

isa/Makefile
ybc-alkaid: Sure, I've updated the Makefile.

aswaterman: Thanks.

PR #648, opened 2026-03-11
RFC: add testcase for new p extension

chihminchao:
The current version is not fully verified with spike implementation because there are some issues in toolchains I can get.


chihminchao:
Update: 2026/05/21
1. squash commits to reduce rebase effort
2. verification the implementation with latest spike (there are some fix which will be sent to riscv-isa-sim later)

To do
1. add vxsat initialization and checking for instructions having saturation behavior.

chihminchao:
Update: 2026/06/03
1. add vxsat checking
2. fix multiplication and shift testcases

aswaterman (approved):

PR #651, opened 2026-04-22
Add hypervisor tests for implicit G-stage faults during VS-stage walks

nadime15:
This PR adds four RV64 hypervisor tests exercising implicit G-stage faults raised during VS-stage page table walks:

- 2-stage_translation_implicit_load_error
- 2-stage_translation_implicit_load_error_hs
- 2-stage_translation_implicit_store_error
- 2-stage_translation_implicit_store_error_hs

The load variants trigger the fault by leaving the G-stage PTE invalid (no PTE_V), causing the implicit PTE read during the VS-stage walk to fault. The store variants omit PTE_W on the G-stage PTE, so G-stage grants read but denies write, causing the implicit PTE store for the A/D writeback to fail. These require hardware A/D updates, which occur when Svadu is enabled via ADUE.

Each test checks mcause/scause, mtval/stval, and mtval2/htval and verifies that mtinst/htinst contains the expected pseudoinstruction:

- 0x3000 — 64-bit implicit VS-stage PTE load
- 0x3020 — 64-bit implicit VS-stage PTE store

### DIFF
--- a/isa/Makefile
+++ b/isa/Makefile
@@ -26,6 +26,7 @@ include $(src_dir)/rv64ssvnapot/Makefrag
 include $(src_dir)/rv64mi/Makefrag
 include $(src_dir)/rv64mzicbo/Makefrag
 include $(src_dir)/hypervisor/Makefrag
+include $(src_dir)/hypervisor-svadu/Makefrag
 else
 include $(src_dir)/rv32ui/Makefrag
 include $(src_dir)/rv32uc/Makefrag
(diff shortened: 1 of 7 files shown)


nadime15:
Test have been tested on Spike as well as on Sail (I work on a hypervisor support for Sail, its not in the main branch yet) however the store cases fail on QEMU (not sure if i did a mistake or QEMU is actually wrong here, I opened a bug report: https://gitlab.com/qemu-project/qemu/-/work_items/3426)

aswaterman:
I suggest moving the tests that require on Svadu into their own directory, e.g. `hypervisor-svadu`.

I would ignore the detail about systems that implement neither Svadu nor Svade.  This is a legacy detail.  For the purposes of these tests, we can assume that the implementation supports Svade (as required by the RVA/RVB profiles) and might optionally support Svadu.

Assuming your description on the QEMU ticket is correct, it does sound like a QEMU bug, so we can go ahead and merge these tests once you've addressed my comments.

nadime15:
Unfortunately, its a bit asymmetric, the load tests don’t depend on Svadu, whereas the store cases do. Do you still recommend moving _all_ of these tests into `hypervisor_svadu`?

aswaterman:
No, I recommend moving only the ones that depend on Svadu to hypervisor-svadu.  (Hyphen rather than underscore, I think.)  It's OK if only load tests are in hypervisor and only store tests are in hypervisor-svadu for now.  We can always add more tests to fill out the cross product at a later date.

nadime15:
Ah sorry, I read too quickly and missed that I only need to move the Svadu dependent tests into hypervisor-svadu. Will do, thanks!

nadime15:
Ping @aswaterman

aswaterman (approved):

PR #658, opened 2026-08-14
hypervisor: align check_fault to 4-byte boundary

i3abghany:
The check_fault label is positioned after a `j fail` instruction, then installed as a trap vector. If built with the C extension (e.g., the one I was experimenting with: `rv64imach_zicsr_zifencei`), the label could land on a 2-byte-aligned address. `stvec.BASE` assumes a 4-byte-aligned address.

The low bits are masked out, and the registered address is one instruction early (the `j fail`).

To reproduce:

```bash
cd riscv-tests/isa
git show HEAD:isa/hypervisor/2-stage_translation_implicit_load_error_hs.S > /tmp/pristine.S

for arch in rv64gh rv64imach_zicsr_zifencei; do
  riscv64-unknown-elf-gcc -march=$arch -mabi=lp64 -static -mcmodel=medany \
    -fvisibility=hidden -nostdlib -nostartfiles -I./../env/p -I./macros/scalar \
    -T./../env/p/link.ld /tmp/pristine.S -o /tmp/hs-$arch.elf
  riscv64-unknown-elf-nm /tmp/hs-$arch.elf | grep ' check_fault$'
  spike --isa=$arch /tmp/hs-$arch.elf; echo "  exit=$?"
done
```

Output:

```bash
0000000080000280 t check_fault
  exit=0
0000000080000206 t check_fault
*** FAILED *** (tohost = 2)
  exit=2
```

Reproduced on:

```bash
$ spike --help
Spike RISC-V ISA Simulator 1.1.1-dev

$ riscv64-unknown-elf-gcc --version
riscv64-unknown-elf-gcc (g6afcc4f6d) 16.1.0
```


aswaterman (approved):
Yep, thanks for the fix.

nadime15:
I just stumbled across this fix. The non-_hs variants introduced in https://github.com/riscv-software-src/riscv-tests/pull/651/changes need this fix too.

Issue #659, opened 2026-09-03
Add Zacas test


vsoriap:
## Summary
Looking for the in the rv64ua I was unable to find any Zacas tests. In the past the #492 tried to introduce such tests, but it was closed before merge.

## Actions
Recover the commits of #492 and cover Zacas extension.

PR #662, opened 2026-09-17
Add Zacas support

vsoriap:
This PR recovers PR #492 which adds tests for the Zacas extension that introduces the following instructions:

- amocas.w for RV32 and RV64
- amocas.d for RV32 and RV64
- amocas.q for RV64 only

The specification is here: https://github.com/riscv/riscv-zacas
The amocas instructions are emitted using a .word directive pending assembler support.


vsoriap:
This also solves #659

aswaterman:
The PR was initially failing because our toolchain is too old.  That's fixed.

But there are still more problems.  Some of the test cases are failing.  Make sure they pass on Spike before resubmitting.

vsoriap:
Thanks @aswaterman, I just took the code from the PR but did not check all of the tests. I've updated the failing ones and double checked them with spike.

aswaterman:
LGTM now, thanks!

aswaterman (approved):

